PR #2, opened 2021-09-27
Feature/orders history

wender:
Adding Orders History support using User's Organization and Cost Center

#### How to test it?

[Workspace](https://profile--sandboxusdev.myvtex.com/account#/orders)

A) User `wender.lima@gmail.com` has access to orders placed under the same Organization
B) User `wender.lima@vtex.com.br` has access to orders placed under the same Cost Center

Both passwords are `Wender@123`

Use case A
image

Use case B
image


vtex.storefront-permissions/configuration.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "Organization",
```
polishq, line 2: When I was testing permissions for orderquote, I had to set the `name` as `vtex.orderquote` in order to get the correct permissions back from the backend `checkUserPermissions` GraphQL query. When the query executes, doesn't it look at the app ID of the sender (without the version number) and then return any features matching that ID?

polishq: Also, I think this permissions configuration and builder should be moved to the frontend app, right?

polishq (approved):

PR #14, opened 2021-12-17
[B2BORG=42] Enable MD immediate indexing and deprecate field

polishq:

Enables newly created organizations, organization requests, cost centers, and changes to any of the above to be immediately visible in their respective UIs.

Also deprecates the `costCenters` array in each organization data entity as this is unnecessary. Instead, to get an organization's cost centers, apps can use the `getCostCentersByOrganizationId` query.

#### How to test it?

Linked in https://quotes--sandboxusdev.myvtex.com


albertm805 (approved):
LGTM

PR #18, opened 2022-02-28
B2BORG-68

arturmagalhaesjr:
- Storefront permissions listUsers query fixed

- Users information is not shown in storefront

#### How to test it?

Credentials:
artur.magalhaes@vtex.com.br / Vtex1234

[https://org--b2bstoreqa.myvtex.com/account#/organization](https://org--b2bstoreqa.myvtex.com/account#/organization)

### DIFF
--- a/node/resolvers/index.ts
+++ b/node/resolvers/index.ts
@@ -140,8 +140,8 @@ export const QUERIES = {
       permissions
     }
   }`,
-  listUsers: `query users($organizationId: ID, $costCenterId: ID, $roleId: ID) {
-    listUsers(organizationId: $organizationId, costCenterId: $costCenterId, roleId: $roleId) @context(provider: "vtex.storefront-permissions") {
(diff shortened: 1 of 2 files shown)


node/resolvers/index.ts
```diff
@@ -140,8 +140,8 @@ export const QUERIES = {
       permissions
     }
   }`,
-  listUsers: `query users($organizationId: ID, $costCenterId: ID, $roleId: ID) {
```
polishq, line 143: I don't think we can remove the `roleId` variable, as it is used by `node/resolvers/message.ts` to get a list of users that have a certain role. Can you explain why this solves the problem? Can you think of an alternative?

polishq (changes requested):

arturmagalhaesjr:
The permissions was outdated, this PR doesn't make sense

PR #22, opened 2022-03-24
[B2BORG-83] Allow all payment methods to be assigned to orgs

polishq:
### What problem is this solving?

Our team recently decided that when assigning payment terms to an organization, any of the store's payment methods should be eligible for assignment. Until now, we have only been allowing Promissory payment types to be assigned. This PR alters the getPaymentTerms query so that all enabled payment methods are available for assignment.

For credit cards (Visa, Mastercard, etc), the app filters these out of the list and adds a generic "Credit card" entry to the array instead. This is because all credit cards in checkout are combined into a single "Credit card" option.

This PR also fixes an unrelated bug where the wrong app's permissions were being checked in the orders route handler. We need to check the user's permissions for vtex.b2b-orders-history and not their permissions for vtex.b2b-organizations.

### How to test it?

New version linked here: https://b2bsuite--sandboxusdev.myvtex.com/
Check the available payment options here: https://b2bsuite--sandboxusdev.myvtex.com/admin/b2b-organizations/organizations/1574d7ea-1250-11ec-82ac-020154316047/

### DIFF
--- a/node/resolvers/index.ts
+++ b/node/resolvers/index.ts
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { ForbiddenError } from '@vtex/api'
 
-import { CONNECTOR } from '../constants'
+// import { CONNECTOR } from '../constants'
 import {
   schemas,
   ORGANIZATION_REQUEST_DATA_ENTITY,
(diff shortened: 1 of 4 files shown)


node/resolvers/index.ts
```diff
@@ -1760,11 +1760,55 @@ export const resolvers = {
...
+              'Elo',
+              'Banricompras',
+              'JCB',
+              'Cabal',
+              'Nativa',
+              'Naranja',
+              'Nevada',
+              'Shopping',
+              'Credz',
+            ]
```
arturmagalhaesjr, line 1799: I would put this constant array outside the filter function to avoid processing each loop and make it clear. However, The best solution would be if this array comes from the app setting, what do you think?

arturmagalhaesjr (changes requested):

arturmagalhaesjr (dismissed):
LGTM

PR #35, opened 2022-05-04
Feature/cy runner parallel

thyarles:

Add parallelization to run Cypress tests.


cy-runner.yml
```diff
@@ -113,20 +109,21 @@ strategy:
...
+      - cypress-shared/integration/b2b/2.1.2*
+
+  orgA:
     enabled: false
     sendDashboard: true
     hardTries: 1
     stopOnFail: false
-    parallel: false
-    runInOrder: true
+    parallel: true
```
Syed-Vtex, line 126: This has to be in sequence

thyarles: Sorry, I didn't understand this one.

Syed-Vtex: We should run them in order
If we run them in parallel then it will fail

PR #36, opened 2022-05-05
[B2BORG-6] Organization storefront UI - Sales admin view

arturmagalhaesjr:
**What problem is this solving?**

It was added into getUsers approach the pagination approach in order to attend the[ [B2BORG-6]](https://github.com/vtex-apps/b2b-organizations/pull/38) requirements.

### DIFF
--- a/node/resolvers/Queries/CostCenters.ts
+++ b/node/resolvers/Queries/CostCenters.ts
@@ -276,7 +276,6 @@ const costCenters = {
       pageSize,
       sortOrder,
       sortedBy,
-      isSalesAdmin,
     }: {
       id: string
       search: string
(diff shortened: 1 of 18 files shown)


polishq (changes requested):

node/resolvers/Queries/CostCenters.ts
```diff
@@ -287,28 +287,49 @@ const costCenters = {
...
-    if (!sessionData?.namespaces['storefront-permissions']) {
-      throw new GraphQLError('organization-data-not-found')
-    }
-
     const {
-      organization: { value: userOrganizationId },
-    } = sessionData.namespaces['storefront-permissions']
+      data: { checkUserPermission },
+    }: any = await storefrontPermissions
+      .checkUserPermission('vtex.b2b-orders-history@0.x')
```
polishq, line 300:
```suggestion
      .checkUserPermission('vtex.b2b-organizations@1.x')
```

node/resolvers/Queries/CostCenters.ts
```diff
@@ -287,28 +287,49 @@ const costCenters = {
...
+            checkUserPermission: null,
+          },
+        }
+      })
 
-    if (!id) {
-      // get user's organization from session
-      id = userOrganizationId
-    }
+    const isSalesAdmin = checkUserPermission?.role.slug.match(/sales-admin/)
```
polishq, line 314: Ideally, rather than checking their role, we should check if they have a specific permission. This would require adding a new permission to b2b-organizations like "View Cost Centers In Any Organization", so perhaps we can put this on the backlog.

node/resolvers/Queries/index.ts
```diff
@@ -102,12 +115,17 @@ const Index = {
...
+      sortOrder,
+      sortedBy,
     }
 
     return storefrontPermissions
       .listUsers(variables)
       .then((result: any) => {
-        return result.data.listUsers
+        return result.data.listUsersPaginated
       })
```
polishq, line 129: The permissions check in this resolver needs to be updated to allow sales admins to view users in any organization. You'll notice that if the `adminUserAuthToken` is present, all of the permissions checks are skipped. Since you are an admin user, that's what's allowing the query to return all users in your workspace. 

Similar to what you did in the Cost Centers query, you should check the user's role. Then you could change the conditional to `if (!adminUserAuthToken && !isSalesAdmin) { ...` . Otherwise the sales admin will only be able to view users that are part of the sales admin's "main" organization.

polishq (changes requested):

arturmagalhaesjr (commented):
done

polishq (dismissed):
LGTM

arturmagalhaesjr (commented):
done

polishq (approved):
LGTM

PR #38, opened 2022-05-10
[B2BORDERS-8] New route and permission checks for orders history

polishq:

Adds a permission check to the app's `order` route to ensure that the frontend user has permission to view the order in question. Also the app now uses the admin version of the OMS API to get the order details, because this API returns the shopper's actual email inside the `clientProfileData` object, instead of a masked version. We need this email to check in `b2b-order-history` if the current user was the one who placed the order.

Also adds a new route for order cancellation requests since normally the frontend user is only able to request the cancellation of an order that they personally placed. This route uses the app's credential (and the `CancelaPedidos` policy) to submit the cancellation request using checkout's private API route, after first ensuring that the user has permissions to do so (i.e. they are part of the same organization as the user who placed the order). 

#### How to test it?

This new version and a new version of `b2b-orders-history` are both linked in http://b2bsuite--sandboxusdev.myvtex.com

Log in as an organization user, go to My Account -> Orders (there are two Orders links, choose the second one) and try viewing and cancelling an order. Also try changing the URL to an order ID that you should not be able to access.

### DIFF
--- a/node/resolvers/Routes/index.ts
+++ b/node/resolvers/Routes/index.ts
@@ -1,4 +1,4 @@
-import { ForbiddenError } from '@vtex/api'
+import { UserInputError, ForbiddenError } from '@vtex/api'
 
 import {
   COST_CENTER_DATA_ENTITY,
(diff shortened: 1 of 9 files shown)


node/resolvers/Routes/index.ts
```diff
@@ -70,12 +70,103 @@ const Index = {
...
+
+        return null
+      })
+
+    const profileEmail = sessionData?.namespaces?.profile?.email?.value
+
+    const checkPermissionAgainstOrder = (
+      permissions: string[],
+      orderData: any
+    ) => {
```
arturmagalhaesjr, line 111: I strongly believe that could be outside of the Index object once this function is like an if wrapper, what do you think?

node/resolvers/Routes/index.ts
```diff
@@ -168,6 +259,36 @@ const Index = {
...
+    } = ctx
+
+    if (!orderId) {
+      throw new UserInputError('Order ID is required')
+    }
+
+    // use order route handler to verify user has access to order
+    Index.order(ctx).catch(() => {
+      throw new ForbiddenError('Access denied')
+    })
```
arturmagalhaesjr, line 279: Is it necessary to call the Index.order and pass through to the entire route only to check the permission? I'd create a new function to check it and call it in both routes. Even though, this throw could be called after the response calls once it's not into a try/catch + async/await statement. What do you think?

arturmagalhaesjr (changes requested):

node/resolvers/Routes/index.ts
polishq: I'll have to refactor the function to take additional arguments, but I agree it's a good idea

node/resolvers/Routes/index.ts
polishq: Sure, it's worth a shot 👍

polishq: Done

node/resolvers/Routes/index.ts
polishq: Done

arturmagalhaesjr (approved):
LGTM

PR #59, opened 2022-07-04
Add changeable timeout for Cypress-Runner

thyarles:

Add timeout option for Cypress-Runner in Dispatch mode.

### DIFF
--- a/.github/workflows/qe-dispatch.yml
+++ b/.github/workflows/qe-dispatch.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   quality-engineering:
     name: QE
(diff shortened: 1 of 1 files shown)


.github/workflows/qe-dispatch.yml
```diff
@@ -6,6 +6,9 @@ on:
       cyRunnerBranch:
         description: 'Which branch to use on Cy-Runner?'
         default: 'main'
+      cyRunnerTimeOut:
+        description: 'Wait how much time (in minutes) for the job to complete?'
+        default: '30'
```
Syed-Vtex, line 11: If Update `default: '30'` to `default: '40'`
Then it will use 40 minutes right

Syed-Vtex (approved):

shashidhar85 (approved):

PR #67, opened 2022-07-19
[B2BORG-133] App Review

arturmagalhaesjr:
refactor: App Review: Added logging, fixed security issues and code cleanup
doc: changelog updated

## App Review

Review the app for each of the following, as appropriate:

Logging optimization

GraphQL mutation security

Admin navigation permissions

See detailed instructions here: https://www.notion.so/vtexhandbook/2022-Q3-App-Review-Checklist-a68487945f864ce0a2b216a43d31efa3

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -79,7 +79,7 @@ type Mutation {
     notes: String
   ): MutationResponse
   deleteOrganizationRequest(id: ID!): MutationResponse
-  createOrganization(input: OrganizationInput!): MasterDataResponse
+  createOrganization(input: OrganizationInput!): OrganizationCostCenterResponse
     @checkAdminAccess
     @cacheControl(scope: PRIVATE)
   createCostCenter(
@@ -284,6 +284,13 @@ type Address {
 
 type MasterDataResponse {
   id: String
+  href: String
+  status: String
+}
+
+type OrganizationCostCenterResponse {
+  id: String
+  costCenterId: String
   href: String
   status: String
 }
--- a/node/utils/constants.ts
+++ b/node/utils/constants.ts
@@ -1,7 +1,12 @@
-export const STATUS = {
+export const ORGANIZATION_STATUSES = {
   ACTIVE: 'active',
+  INACTIVE: 'inactive',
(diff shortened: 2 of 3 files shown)


graphql/schema.graphql
```diff
@@ -93,19 +97,35 @@ type Mutation {
     paymentTerms: [PaymentTermInput]
     priceTables: [String]
   ): MutationResponse
+    @withSession
+    @withPermissions
```
polishq, line 101: Only store admins should be able to update an organization, so these directives can be removed and replaced with a new `@checkAdminAccess` directive that checks the `adminUserAuthToken`

Same with `createOrganization` above and `deleteOrganization` below

polishq (changes requested):

graphql/schema.graphql
polishq: Additionally, `@withPermissions` doesn't actually control access, it just gives the resolver access to permission data. There are many mutations (for example all of the Cost Center mutations) where the app is not checking if the user has permission to do that particular operation.

For these cases where the mutation can be called by either a store user or admin user, we can create a directive that checks for either a `adminUserAuthToken` or a `storeUserAuthToken` and validates whichever one is present. Maybe `@checkUserAccess`, what do you think?

Then we should still add some additional validation in each resolver to make sure (for example) the user has access to the organization they're trying to create a cost center in.

node/utils/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const STATUS = {
+  ACTIVE: 'active',
+  APPROVED: 'approved',
+  DECLINED: 'declined',
+  PENDING: 'pending',
+  SUCCESS: 'success',
+}
```
polishq, line 7: I suggest separating this into two constants, one for Organization Request statuses (PENDING, APPROVED, DECLINED) and one for Organization statuses (ACTIVE, INACTIVE, ON-HOLD)

Then for example in the `updateOrganization` resolver we could check if the provided status argument is a valid value

polishq (changes requested):

polishq (approved):
LGTM

PR #73, opened 2022-09-01
[CUSTOMERS-4] User one2many admin management

arturmagalhaesjr:

Added a graphql query to fetch all organizations from a user by email parameter

It's a part of one to many feature

#### How to test it?

[https://b2borg2--sandboxusdev.myvtex.com/admin/graphql-ide](https://b2borg2--sandboxusdev.myvtex.com/admin/graphql-ide)

```graphql 
query organizations {
  getOrganizationsByEmail(email:"artur.magalhaes@vtex.com.br") {
    costId
    orgId
    roleId
    organizationName
    costCenterName
    role {
      name
    }
  }
}
```

Depends on: [https://github.com/vtex-apps/storefront-permissions/pull/57](https://github.com/vtex-apps/storefront-permissions/pull/57)

### DIFF
--- a/node/resolvers/Queries/Organizations.ts
+++ b/node/resolvers/Queries/Organizations.ts
@@ -134,7 +134,7 @@ const Organizations = {
 
     if (
       !adminUserAuthToken &&
-      !checkUserPermission?.role.slug.match(/sales-admin/)
+      !checkUserPermission?.permissions.includes('add-sales-users-all')
     ) {
       const sessionData = await session
         .getSession(sessionToken as string, ['*'])
@@ -150,7 +150,7 @@ const Organizations = {
           return null
         })
 
-      if (checkUserPermission?.role.slug.match(/customer-admin/)) {
+      if (checkUserPermission?.permissions.includes('add-users-organization')) {
         const orgId =
           sessionData?.namespaces?.['storefront-permissions']?.organization
             ?.value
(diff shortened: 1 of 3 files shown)


node/resolvers/Queries/Organizations.ts
```diff
@@ -105,6 +105,23 @@ const Organizations = {
     }
   },
 
+  getOrganizationsByEmail: async (
+    _: void,
+    { email }: { email: string },
+    { clients: { storefrontPermissions }, vtex: { logger } }: any
```
polishq, line 111:
```suggestion
    { clients: { storefrontPermissions }, vtex: { logger } }: Context
```

node/resolvers/Queries/Organizations.ts
```diff
@@ -105,6 +105,23 @@ const Organizations = {
     }
   },
 
+  getOrganizationsByEmail: async (
+    _: void,
+    { email }: { email: string },
```
polishq, line 110: Rather than allowing any email to be provided as an argument, what if the query had no arguments and the query automatically extracted the email from the current user's session? Is there any situation where one user would need to query another user's organizations?

polishq (changes requested):

node/resolvers/Queries/Organizations.ts
polishq: If it's a valid scenario, we need to add a permissions check to see if the current user has permission to view another user's organization data

polishq (approved):
LGTM

PR #75, opened 2022-09-14
[PERMISSION-32] Challenge check for active Organizations and active Users

arturmagalhaesjr:

When checking the user permission, only keep checking if the organization is active

#### How to test it?

https://b2borg2--sandboxusdev.myvtex.com/account#/organization

Was added a new graphql query:
```graphql
query {
  checkOrganizationIsActive(id:"16017bc2-307f-11ed-835d-122119b02651")
}
```


btalma (approved):
LGTM

btalma (approved):

PR #79, opened 2022-09-27
[B2BORG-141]  Organizations dropdown selector

arturmagalhaesjr:
- Updated graphql schema and added and fix a bug when the query is called

Interface block to list all Organizations, when selected this will set the user session with the current selection, applying all customizations related to the organization

Inactive Organizations won’t be selectable (grayed out)
(also throw error in mutation resolver if inactive organization is provided as argument)

Related PRs:

https://github.com/vtex-apps/storefront-permissions/pull/59
https://github.com/vtex-apps/b2b-organizations/pull/70

### DIFF
--- a/node/resolvers/Queries/Users.ts
+++ b/node/resolvers/Queries/Users.ts
@@ -376,7 +376,7 @@ const Users = {
     const {
       clients: { catalog },
       vtex: { logger },
-    } = ctx as any
+    } = ctx
 
     let access = false
     let availableSalesChannels: any = {}


node/resolvers/Queries/Users.ts
```diff
@@ -376,7 +376,7 @@ const Users = {
     const {
       clients: { catalog },
       vtex: { logger },
-    } = ctx
+    } = ctx as any
```
polishq, line 379: You shouldn't need the `as any` here

polishq (changes requested):

polishq (approved):
LGTM

PR #91, opened 2023-01-24
fix: remove json webtoken dependecy

arturmagalhaesjr:

fix: remove json webtoken dependecy


polishq (approved):
LGTM

PR #103, opened 2023-03-31
Feature/search cost center

arturmagalhaesjr:

- Added the businessDocument as search field


albertm805 (approved):
LGTM

PR #110, opened 2023-05-30
Hotfix - persist state registration on organization approval

Rudge:
#### How to test it?

[Workspace](Link goes here!)


node/resolvers/Mutations/Organizations.ts
```diff
@@ -456,13 +457,13 @@ const Organizations = {
                   ...(defaultCostCenter.customFields && {
                     customFields: defaultCostCenter.customFields,
                   }),
+                  stateRegistration: defaultCostCenter.stateRegistration,
```
AnnaChiu95, line 460: So the main fix was to include `stateRegistration` in the `updateOrganizationRequest` mutation when creating a cost center and add unit tests to check that it's there?

`stateRegistration` is optional like `businessDocument` and it's value can be `null` so the pattern to include them was to check for the value first (see below), but I'm not seeing any issues when simply doing `stateRegistration: defaultCostCenter.stateRegistration`. 🤔  In both master and dev workspaces, it looks like `stateRegistration` is saved as an empty string instead of `null` when it's left blank.

```
...(defaultCostCenter.businessDocument && {
    businessDocument: defaultCostCenter.businessDocument,
})
```

node/package.json
```diff
@@ -12,23 +12,25 @@
...
     "vtex.storefront-permissions": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.storefront-permissions@1.33.2/public/@types/vtex.storefront-permissions"
   },
+  "resolutions": {
+    "@types/testing-library__dom": "6.12.1",
+    "@types/express": "4.16.0",
+    "@types/express-serve-static-core": "4.16.0"
+  },
   "scripts": {
-    "lint": "tsc --noEmit && tslint -c tslint.json './**/*.ts'"
+    "lint": "tsc --noEmit --pretty",
```
mairatma, line 33: Why remove tslint? Now this script only builds the code, but doesn't run lint anymore.

node/package.json
```diff
@@ -12,23 +12,25 @@
...
+    "casual": "^1.6.2",
+    "ts-mockito": "^2.6.1",
     "typescript": "3.9.7",
     "vtex.storefront-permissions": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.storefront-permissions@1.33.2/public/@types/vtex.storefront-permissions"
   },
+  "resolutions": {
+    "@types/testing-library__dom": "6.12.1",
+    "@types/express": "4.16.0",
+    "@types/express-serve-static-core": "4.16.0"
+  },
```
mairatma, line 31: Can you explain why you needed these? What kind of errors were you getting? Just trying to see if we can do without.

node/resolvers/Mutations/Organizations.test.ts
```diff
@@ -0,0 +1,144 @@
...
+import {
+  anything,
+  capture,
+  instance,
+  mock,
+  reset,
+  spy,
+  verify,
+  when,
+} from 'ts-mockito'
```
mairatma, line 11: We usually just use jest directly at VTEX for [mocking](https://jestjs.io/docs/mock-functions), it's the most popular tool out there, and the most complete as well. Plus, we're already using jest to run the tests, so it's one less dependency.

Let me know if you had trouble while using jest for mocking though, and we can either discuss here or setup some time for a pair programming where I can show these test tools for node.js/js

node/.vtexignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
```
mairatma, line 1: I believe this file is unnecessary, the root `.vtexignore` file should already cover this.

node/babel.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
mairatma, line 1: This is usually not required, unless we need very specific custom babel configs, which doesn't seem the case here. Do you know why you had to add this file?

node/resolvers/config.ts
```diff
@@ -19,121 +19,123 @@ export const getAppId = (): string => {
   return appName
 }
 
-const checkConfig = async (ctx: Context) => {
-  const {
-    vtex: { logger },
-    clients: { mail, masterdata, vbase },
-  } = ctx
+class Config {
```
mairatma, line 22: In JS, functions are generally preferred to classes, except in a few cases where having instances really helps. In this case, we could just keep the function like before.

My suspicion is that you changed this to a class because of troubles mocking the `checkConfig` function, right? When using jest mocks this is feasible though, you can use the [`jest.mock` function](https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options) in your test file, which will automatically make all imports for this file receive a mocked version of it instead.

So, in your test file you would do something like this:

```ts
import checkConfig from '../config'

jest.mock('../config')

// The following line is just so you can access jest.fn functions on `mockedCheckConfig`, otherwise
// Typescript will think it's the real checkConfig function instead.
const mockedCheckConfig = checkConfig as ReturnType<typeof jest.fn>

// In your test code you can mock the return value
mockedCheckConfig.mockReturnValue('mockedResult')
// Or the whole implementation
mockedCheckConfig.mockImplementation((param1, param2) => { // any logic at all here})
```

node/resolvers/Mutations/Organizations.ts
```diff
@@ -337,12 +337,12 @@ const Organizations = {
       await masterdata.updatePartialDocument({
         dataEntity: ORGANIZATION_DATA_ENTITY,
         fields: {
-          name,
-          ...((tradeName || tradeName === '') && { tradeName }),
           collections,
+          ...((tradeName || tradeName === '') && { tradeName }),
+          customFields,
+          name,
```
mairatma, line 343: Are these order changes here required? How do they affect the solution?

node/resolvers/Mutations/Organizations.ts
mairatma: Good point, @AnnaChiu95, It's probably safer not do send the value at all if it's not set, just to be sure.

node/resolvers/Mutations/Organizations.ts
mairatma: Ahhhh, I see, you're sorting alphabetically, right? I like it, it's just been so long since I've seen someone do this. I'm all for it! 👏  We can later look into a lint rule to enforce this, if the rest of the team also likes it 😬 .

node/resolvers/Mutations/Organizations.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    beforeAll(async () => {
+      result = await Organizations.updateOrganizationRequest(
+        instance(mock()),
+        input,
+        mockedContext
+      )
+    })
+
+    it('should return oranization id with success', () => {
```
mairatma, line 121: Small typo:

```suggestion
    it('should return organization id with success', () => {
```

node/resolvers/Mutations/Organizations.test.ts
```diff
@@ -0,0 +1,144 @@
...
+import { ORGANIZATION_REQUEST_STATUSES } from '../../utils/constants'
+import type { Settings } from '../config'
+import Config from '../config'
+import B2BSettings from '../Queries/Settings'
+import Organizations from './Organizations'
+
+const orgId = casual.uuid
+const stateRegistration = casual.word
+const spyOrganizations = spy(Organizations)
+const spyB2BSettings = spy(B2BSettings)
```
mairatma, line 23: Jest can do this via the `jest.mock` function as well:

```ts
jest.mock('../Queries/Settings'')
jest.mock('./Organizations')
```

node/resolvers/Mutations/Organizations.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    it('should return oranization id with success', () => {
+      expect(result.status).toEqual('success')
+      expect(result.id).toEqual(orgId)
+    })
+
+    it('should call create organization with data', () => {
+      verify(
+        spyOrganizations.createOrganization(anything(), anything(), anything())
+      ).once()
```
mairatma, line 129: With jest this would be something like this:

```ts
expect(OrganizationsMock.getB2BSettings).toHaveBeenCalledTimes(1)
```

You can also check for params, but since you were passing everything() here, then you can just omit them to verify that the function was called.

mairatma (commented):
Thanks a lot for this fix @Rudge ! 👏 ✨ 

I've added some comments, but most are just about frontend testing patterns using jest. Let me know what you think!

.vtexignore
```diff
@@ -3,3 +3,4 @@ node_modules/
 .eslintrc
 .gitignore
 README.md
+./**/*.test.ts
```
mairatma, line 6: You mentioned you were having problems with this not working to ignore test files in VTEX IO's build. I also had trouble with this the last time I set up a VTEX IO repo with tests, and the way it worked for me was:

```suggestion
**/*.test.*
```

You can see the whole .vtexignore file [here](https://github.com/vtex/buyer-organizations-supplier/pull/40/files#diff-493b51e38bdabe924ce34da582e5e8345ff33831c58186f0cb03e75099b970d7R16) as well in case you have trouble with other types of files too.

node/resolvers/Mutations/Organizations.ts
Rudge: Yes @AnnaChiu95 the change just was it. The another is because of sorting alphabetically the parameter, I got in my IDE and fix it.

I can make a test to validate this point, but looking the [`createOrganization`](https://github.com/vtex-apps/b2b-organizations-graphql/blob/hotfix/B2BTEAM-1191-state-registration/node/resolvers/Mutations/Organizations.ts#L86) implementation, the parameter is validated, so it is unnecessary to do again.

node/.vtexignore
Rudge: I will remove this file, I put just to try to pass the [vtexio/build](https://master--vtex.myvtex.com/admin/builder-log/build-log-viewer/8db2adce332948cea5381ddc09105d6f). But didn't works...

.vtexignore
Rudge: OK, let me change. I found it strange because locally it is works to link in workspace.

node/package.json
Rudge: I had some troubles with the Typescript compilation on lint execution, mainly for the `@types/express-serve-static-core`, so I checked the template and there is this configuration [there](https://github.com/vtex-apps/service-example/blob/main/node/package.json#L16).

node/package.json
Rudge: Oops, I was fixing the issues with lint and I took the [template example](https://github.com/vtex-apps/service-example/blob/main/node/package.json#L22).

node/babel.config.js
mairatma: I see! I believe I also got this kind of problem before, it's something with the tsconfig configuration not knowing that esmodules should be used. My guess is that if you set `esModuleInterop: true` is node/tsconfig.js it should work without this file. Can you try it out? For the other repo I setup tests for, this line is [there](https://github.com/vtex/buyer-organizations-supplier/blob/main/node/tsconfig.json), so it may be what makes it work there and not here.

node/package.json
mairatma: Got it, probably due to the craziness of VTEX IO's outdated deps... tslint is also something not much used nowadays, people tend to use eslint directly. But let's not touch this now. It's fine to add this here to make it work now.

node/package.json
mairatma: Maybe tslint is already integrated into typescript itself, but I don't think so. Can you double check if when you run this it catches lint errors (maybe fake one and see if it's detected like this vs by calling tslint directly).

Rudge: @mairatma. Why is the lint configured different in root [`package.json`](https://github.com/vtex-apps/b2b-organizations-graphql/blob/master/package.json#L6) and node [`package.json`](https://github.com/vtex-apps/b2b-organizations-graphql/blob/master/node/package.json#L32)?

node/babel.config.js
Rudge: I put the config but didn't work... :-(

mairatma: Don't worry, then. Let's leave this babel file here if it's working. We'll optimize these setups later, they're really complex right now :(

node/package.json
Rudge: Now it isn't failing, so I keep the previous config to decrease changes in this PR.

node/resolvers/config.ts
Rudge: I tried to rollback the implementation of `checkConfig` and use the mock, but i failure in all attempts...

Rudge:
@arturmagalhaesjr and @wender I opened this PR to fix the problem with `stateRegistration` but I had some problems with the merge, when I pulled the changes that you made, the fix was made in [another PR](https://github.com/vtex-apps/b2b-organizations-graphql/commit/12e8d5cab077d2034d0ce26266985816f62c79b9#diff-ac25af63931d8c82c74361072a12cf08059bdb5a75d58f7fc1586f9d9cb91ef8R504). So isn't this PR more necessary? CC: @mairatma

mairatma:
> So isn't this PR more necessary? 

Oh wow, you're right, it seems like it was fixed there. In that case, we can just keep this PR for adding some tests around it. I'll help you with the Jest mocking tools in the meeting we have later today.

node/mutations/addUser.ts
```diff
@@ -2,7 +2,7 @@ import { print } from 'graphql'
 import gql from 'graphql-tag'
 
 export default print(gql`
-  mutation(
+  mutation (
```
Rudge, line 5: Unfortunately on the lasts PR build, I was forced to execute `eslint --ext js,jsx,ts,tsx . --fix` to pass the checks.

mairatma (approved):
Lgtm, just a few comments on improving the tests by replacing `jest.spyOn` with the functions already existing in the mocked objects/functions via `jest.mock`.

Thanks for working on this @Rudge ! 👏

node/resolvers/Mutations/Organizations.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    mockedContext = mockContext()
+    jest
+      .spyOn(mockedOrganizations, 'createOrganization')
+      .mockImplementation()
+      .mockResolvedValueOnce({
+        costCenterId: casual.uuid,
+        href: '',
+        id: orgId,
+        status: '',
+      })
```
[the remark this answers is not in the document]
Rudge: I tried but didn't work. Let's try one more time.

mairatma (dismissed):
Thanks, @Rudge , looks great!

enzomerca (dismissed):

mairatma (approved):
It was already approved, but approving again! 🚀

enzomerca (approved):

PR #113, opened 2023-06-17
chore: configure Cabueta to report vulnerabilities

Rudge:

Add [Cabueta](https://github.com/vtex/cabueta) action to scan for vulnerabilities.

The idea is call graphql operations with examples using the platform [Nuclei](https://github.com/projectdiscovery/nuclei).

#### How to test it?

After the pull request merge, it will show the workflow to execute on [Actions](https://github.com/vtex-apps/b2b-organizations-graphql/actions)

#### How does this PR make you feel? [:link:](http://giphy.com/)

image

### DIFF
--- a/.github/workflows/cabueta.yml
+++ b/.github/workflows/cabueta.yml
@@ -1,0 +1,18 @@
+name: cabueta
+on:
+  workflow_dispatch:
+
+jobs:
+  cabueta:
+    uses: vtex/cabueta/.github/workflows/cabueta.yml@main
+    secrets:
+      target_url: https://www.vtexfaststore.com/
+    with:
+      dast-check: true
+      target-url: https://master--vtex.myvtex.com/_v/private/graphql/v1
+  print:
+    runs-on: ubuntu-latest
+    needs: cabueta
+    steps:
+      - name: Print output
+        run: echo ${{ needs.cabueta.outputs.report }}
--- a/.github/workflows/cabueta.yml
+++ b/.github/workflows/cabueta.yml
@@ -5,8 +5,6 @@ on:
 jobs:
   cabueta:
     uses: vtex/cabueta/.github/workflows/cabueta.yml@main
-    secrets:
-      target_url: https://www.vtexfaststore.com/
     with:
       dast-check: true
       target-url: https://master--vtex.myvtex.com/_v/private/graphql/v1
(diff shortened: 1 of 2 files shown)


enzomerca (dismissed):

.github/workflows/cabueta.yml
```diff
@@ -0,0 +1,16 @@
+name: cabueta
+on:
+  workflow_dispatch:
+
+jobs:
+  cabueta:
+    uses: vtex/cabueta/.github/workflows/cabueta.yml@main
+    with:
+      dast-check: true
+      target-url: https://master--vtex.myvtex.com/_v/private/graphql/v1
```
Rudge, line 10: @mairatma Do you know if the app is installed on pull request and we can call this environment to scan? I noted that [building](https://master--vtex.myvtex.com/admin/builder-log/build-log-viewer/eec22f3e7eed4f19bbbaa7e614d20d8e) happens.

mairatma: @Rudge Nope, it's not installed on pull request. It would be hard to avoid conflicts if that was the case. Pull requests only build the code today, but they don't publish versions of the app automatically nor create previews we can use here.

I don't think there's a way to have a URL to test automatically on each pull request today, unless we added a new CI/CD pipeline before this one to automatically create a new workspace on a test account for each PR, and then link the app. The linking process is also flaky, so even if we built that CI/CD it would probably fail half of the times 🤦‍♀️ .

mairatma: Also, what would cabueta do with this target-url? Just calling it directly won't access the code for this app, as this url is the same for all GraphQL requests throughout VTEX. To check our APIs it would need to also include valid GraphQL queries in the POST body.

Rudge: The new workflow is execute manually, so it isn't executed on each pull request but it is important to have the environment to scan the APIs. About the graphQL bodies, let me check with the sec team how it is works with these requirements.

mairatma: @Rudge since it doesn't need to be executed in each PR I'd use a test account that has this app installed. The vtex account doesn't have all apps installed in it. You can use account `b2bstoreqa` here, so it would be `https://b2bstoreqa.myvtex.com/_v/private/graphql/v1`. 

After that all that's left is to understand how this tool works for APIs that require body content and which use cases from this app we want to scan with it.

PR #114, opened 2023-06-20
Use the store user token for oms calls

mairatma:

The orders history page is not loading on the production domain of our stores. The call to /b2b/oms/users/orders fails with a 403. More details [here](https://vtex.slack.com/archives/C054PQH3GJF/p1685456760248619).

From my investigation, the store user token works with the failing call to the OMS API, and it's always present on the production domains, while auth token may be different. We also have other logic within b2b suite that gives precedence to the store user token vs auth token, probably for the same reason. So my fix here is the same: give precedence to the store user token, using it when present.

#### How to test it?

Unfortunately this is only really testable on a production domain. So we'll try installing this in a real account to validate the fix before publishing.

For now my test was to make sure the orders history page still loads (check the workspace link below) and that calling the failing OMS API with the production domain's store user cookie also works instead of throwing the 403 it throws now.

[Workspace](https://maira--riviera.myvtex.com/account#/orders-history)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix auth problem when fetching orders history
+
 ## [0.33.2] - 2023-06-09
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


enzomerca (approved):

mairatma:
FYI, @Rudge is talking to the CICD folds to fix the Sonar run. It seems like the token for this repo has expired.

PR #120, opened 2023-07-25
Update roleSlug

rbussola:

Updating roleSlug to get the proper information as it was failing in the final domain with the message 'storefrontPermissions.getRole is not a function'

#### How to test it?

Access https://www.rivieraworks.ro/account?workspace=tkt865094#/organization and try to impersonate any user

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- Update roleSlug
+- Fixed error when fetching a user's role slug
+
+## [0.35.1] - 2023-08-01
+
+### Fixed
+
+- Fix auth problem when fetching orders history
 
 ## [0.35.0] - 2023-07-24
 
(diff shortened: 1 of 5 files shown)


node/resolvers/Mutations/Users.ts
```diff
@@ -170,8 +170,10 @@ const Users = {
         throw new GraphQLError('organization-data-not-found')
       }
 
-      const roleSlug = await storefrontPermissions.getRole(user.roleId)?.data
-        ?.getRole?.slug
+      const roleSlug = await storefrontPermissionsClient.getRole(user.roleId)
+        .then((result: any) => {
+          return result?.data?.getRole?.slug
+        })
```
mairatma, line 176: Thanks, can you just update to avoid using `.then`, and just use await instead? It would be something like this:

```suggestion
      const result = await storefrontPermissionsClient.getRole(user.roleId)
      const roleSlug = result?.data?.getRole?.slug
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Update roleSlug
```
mairatma, line 12: Can you be clearer here? The idea is to explain the change in a few words, so it could be something like:

```suggestion
- Fixed an error happening when fetching a user's role slug.
```

mairatma (approved):
Thanks a lot!

PR #121, opened 2023-07-31
Add impersonate user metrics

mtakeda:

Send events to Analytics Redshift when a user impersonates another user.
Currently, there are two functions used to impersonate and each one will send one different metric (with similar fields) in order to keep tracking about how they are used - there is a comment in the code telling to deprecate one of them, but both are still used)

#### How to test it?
 On [organization manager page](https://takeda--b2bstore005.myvtex.com/account#/organization) select a user to impersonate. After impersonate, click to stop impersonate. Events should appear in Analytics in one hour or so, into "vtex"."schemaless"."b2b_suite_buyerorg_data_raw" schema.

[Workspace](https://takeda--b2bstore005.myvtex.com/account#/organization)

image

image

Types:
```
type ImpersonateUserMetric = {
  kind: 'impersonate-user-graphql-event'
  description: 'Impersonate User Action - Graphql'
}

type ImpersonateB2BUserMetric = {
  kind: 'impersonate-b2b-user-graphql-event'
  description: 'Impersonate B2B User Action - Graphql'
}

export type Metric = {
  name: 'b2b-suite-buyerorg-data'
  account: string
} & (ImpersonateUserMetric | ImpersonateB2BUserMetric)


type ImpersonatePerson = {
  email: string
  buy_org_id: string
  cost_center_id: string
  person_id: string
}

type ImpersonateFieldsMetric = {
  user?: ImpersonatePerson
  target: ImpersonatePerson
  date: string
}

type ImpersonateMetric = Metric & { fields: ImpersonateFieldsMetric }
```

Example of queries and results:

```
SELECT
    payload.account,
    payload.kind,
    payload.fields.date,
    payload.fields.target.email as target_email,
    payload.fields.target.buy_org_id as target_buy_org,
    payload.fields.target.cost_center_id as target_cc

FROM "vtex"."schemaless"."b2b_suite_buyerorg_data_raw"
WHERE payload.name = 'b2b-suite-buyerorg-data'
and payload.kind in ('impersonate-user-graphql-event')
order by ingestion_time desc
```

image

```
SELECT
    payload.account,
    payload.kind,
    payload.fields.date,
    payload.fields.user.email as user_email,
    payload.fields.user.buy_org_id as user_buy_org,
    payload.fields.user.cost_center_id as user_cc,
    payload.fields.target.email as target_email,
    payload.fields.target.buy_org_id as target_buy_org,
    payload.fields.target.cost_center_id as target_cc

FROM "vtex"."schemaless"."b2b_suite_buyerorg_data_raw"
WHERE payload.name = 'b2b-suite-buyerorg-data'
and payload.kind in ('impersonate-b2b-user-graphql-event')
order by ingestion_time desc
```
image


node/resolvers/Mutations/Users.ts
```diff
@@ -612,3 +642,45 @@ const Users = {
...
+  accountName: string,
+  storefrontPermissions: {
+    costcenter: { value: string }
+    userId: { value: string }
+    organization: { value: string }
+    storeUserEmail: { value: string }
+  }
+) {
+  const result = await storefrontPermissionsClient
+    .impersonateUser({ id })
```
mairatma, line 659: The name of the field that this `impersonateUser` receives is `userId`, not `id`, right? Can you fix and make sure that you test this flow after the change so we're sure it's not introducing any problems?

node/resolvers/Mutations/Users.ts
```diff
@@ -612,3 +642,45 @@ const Users = {
...
+    .catch((error: any) => {
+      logger.error({
+        error,
+        message: 'impersonateUser-impersonateUserError',
+      })
+
+      return { status: 'error', message: error }
```
mairatma, line 666: I believe we don't want to send the metrics if the call to impersonate fails, right? Or, if we do want to send metrics in error cases, we need to fix the `sendImpersonateB2BUserMetric` call further above to stop ignoring errors too.

node/resolvers/Mutations/Users.ts
```diff
@@ -612,3 +642,45 @@ const Users = {
...
+    target: {
+      costCenterId: costcenter.value,
+      organizationId: organization.value,
+      email: storeUserEmail.value,
+      id: userId.value,
+    },
+    user: undefined, // no information about original user at this point
+  }
+
+  await sendImpersonateUserMetric(metricParams)
```
mairatma, line 683: This shouldn't be awaited, right? It should run on the background.

mairatma (commented):
It seems like there are two separate flows for impersonation here. Which of these was tested via the instructions you described for this PR? We should make sure to test both use cases, so can you add the instructions for testing both, and make sure they're still working as before?

mtakeda:
> It seems like there are two separate flows for impersonation here. Which of these was tested via the instructions you described for this PR? We should make sure to test both use cases, so can you add the instructions for testing both, and make sure they're still working as before?


There is one function/code tree for impersonation and one for stopping impersonation. The instructions for testing already cover both (starting / stopping impersonation). 
I'm not sure why there are those two functions doing almost the same thing. There is a [comment](https://github.com/vtex-apps/b2b-organizations-graphql/blob/master/graphql/schema.graphql#L211) saying one of them is deprecated, but I was not confident to remove/replace it at all.

node/resolvers/Mutations/Users.ts
mtakeda: Good catch!
I've changed and tested it again. I've noticed that it was working because, in the stopping impersonate flow, it sends null anyway :man_shrugging:

node/resolvers/Mutations/Users.ts
mtakeda: Right again.
Fix to not send metrics to analytics on impersonate errors

mairatma:
> I'm not sure why there are those two functions doing almost the same thing. There is a [comment](https://github.com/vtex-apps/b2b-organizations-graphql/blob/master/graphql/schema.graphql#L211) saying one of them is deprecated, but I was not confident to remove/replace it at all.

Can you try calling the deprecated graphql query directly, to be sure it doesn't fail? Deprecated queries still work, and it might still be used, like you said. The metrics will tell us later, but it's important to run this changed code at least once before merging it. If the UI doesn't call it anymore, it needs to be called directly for testing.

node/resolvers/Mutations/Users.ts
mtakeda: Right. Already fixed when looking another comment

mtakeda:
> > I'm not sure why there are those two functions doing almost the same thing. There is a [comment](https://github.com/vtex-apps/b2b-organizations-graphql/blob/master/graphql/schema.graphql#L211) saying one of them is deprecated, but I was not confident to remove/replace it at all.
> 
> Can you try calling the deprecated graphql query directly, to be sure it doesn't fail? Deprecated queries still work, and it might still be used, like you said. The metrics will tell us later, but it's important to run this changed code at least once before merging it. If the UI doesn't call it anymore, it needs to be called directly for testing.


Actually, frontend calls both, but it could call just one in both scenarios (instead of calling a different one for each scenario). I believe someone marked one of them as deprecated but did not remove it exactly because the front is still calling them.

node/resolvers/Mutations/Users.ts
```diff
@@ -612,3 +642,45 @@ const Users = {
...
+
+  const metricParams: ImpersonateMetricParams = {
+    account: accountName,
+    target: {
+      costCenterId: costcenter.value,
+      organizationId: organization.value,
+      email: storeUserEmail.value,
+      id: userId.value,
+    },
+    user: undefined, // no information about original user at this point
```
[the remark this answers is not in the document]
mairatma: Crazyness 🤯

mairatma (approved):
Lgtm! I'm just not 100% sure if you managed to run the code for both metric types yet. If so, it's good to go. Otherwise, just make sure to run both so we can be sure they're running as expected.

mtakeda:
> Lgtm! I'm just not 100% sure if you managed to run the code for both metric types yet. If so, it's good to go. Otherwise, just make sure to run both so we can be sure they're running as expected.

Yes, I did. 
These are sample results showing both events in Analytics schema

image

mairatma (approved):

PR #125, opened 2023-09-14
Add Orders client and use it to get orders info

enzomerca:

Use Orders endpoint instead of OMS endpoint to obtain order information. The new Orders endpoint returns the correct orderGroup (unlike the OMS endpoint) and this orderGroup is used to generate a new order when clicking on the "Order Again" button.

Ultimately this PR is fixing the KI: "Order Again from B2B Orders History doesn't work".

#### How to test it?

1. Make an order
2. Go to "My Account" -> "Orders"
3. Click on "Order again" button for the order you just created. It should create a new cart with the same items.

[Workspace](https://b2bteam1187--b2bstoreqa.myvtex.com/account#/orders-history)

8EC6DDBE-60AB-405A-9C65-5F0CC9C6C4E2

* https://github.com/vtex-apps/b2b-organizations-graphql/pull/124

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Use Orders endpoint instead of OMS to obtain order information
+
 ## [0.36.0] - 2023-08-11
 
 ### Added
--- a/node/clients/Orders.ts
+++ b/node/clients/Orders.ts
@@ -1,0 +1,35 @@
+import type { InstanceOptions, IOContext, RequestConfig } from '@vtex/api'
+import { JanusClient } from '@vtex/api'
+
+import { statusToError } from '../utils'
+
+export default class OrdersClient extends JanusClient {
+  constructor(ctx: IOContext, options?: InstanceOptions) {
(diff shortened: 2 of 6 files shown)


mairatma (approved):
Code lgtm, the only concern I have here is any differences between the two endpoints, like a missing field on the endpoint we're going to use now. How sure are we that they're compatible for the use cases here?

enzomerca:
> Code lgtm, the only concern I have here is any differences between the two endpoints, like a missing field on the endpoint we're going to use now. How sure are we that they're compatible for the use cases here?

So, I compared the 2 endpoints here: https://vtex-dev.atlassian.net/browse/B2BTEAM-1376
And theoretically we are handling the only problematic difference between the 2 endpoints (the `marketplaceItems` field that can return `null` instead of `[]`).

But yeah, is hard to say 100% that nothing will break. We don't really know if some array field that returns data in our tests, can return `[]` in some other use case but the new API will return `null` and break something. To avoid that we would need to do the same thing that we did for `marketplaceItems` (check if it is `null`, and if it is, set to `[]`) for all other array fields in the response.

What do you think?

mairatma:
> PS: I talked to order-management team and they suggested this approach for now (I can give more details offline)

I was going to suggest talking to them, but you already did! Lgtm, then!

PR #126, opened 2023-09-15
Add event trigger to deleteUser resolver

rafael-marques:

Using Master Data V2, we don't have a trigger to run on "delete event" like in V1. This way we can't run scripts when deleting a user. Only on adding or updating. My motivation for this is to add the ability to run scripts after a delete action.

This modification will trigger an event sending a payload containing the is and email of the deleted user. 
With this, we can add an event listener and run scripts needed for that scenario

#### How to test it?

To test, link this branch to a workspace. Then from another app, listen to an event called `b2b-organizations-graphql.removeUser`

Like this: `service.json`
image

Screenshot 2023-09-15 at 12 00 11 PM

### DIFF
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -412,8 +412,8 @@ const Users = {
         id,
         userId,
       })
-      .then(async (result: any) => {
-        await events.sendEvent('', 'b2b-organizations-graphql.removeUser', {
+      .then((result: any) => {
+        events.sendEvent('', 'b2b-organizations-graphql.removeUser', {
           id,
           email,
         })
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -413,7 +413,7 @@ const Users = {
         userId,
       })
       .then(async (result: any) => {
-        await new Promise(events.sendEvent('', 'b2b-organizations-graphql.removeUser', { id, email }))
+        await events.sendEvent('', 'b2b-organizations-graphql.removeUser', { id, email })
         return result.data.deleteUser
       })
       .catch((error: any) => {
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -379,7 +379,7 @@ const Users = {
     ctx: Context
   ) => {
     const {
-      clients: { storefrontPermissions: storefrontPermissionsClient },
+      clients: { events, storefrontPermissions: storefrontPermissionsClient },
       vtex: { adminUserAuthToken, logger, sessionData, storefrontPermissions },
     } = ctx as Context | any
 
@@ -412,7 +412,8 @@ const Users = {
         id,
         userId,
       })
-      .then((result: any) => {
+      .then(async (result: any) => {
+        await new Promise(events.sendEvent('', 'b2b-organizations-graphql.removeUser', { id, email }))
         return result.data.deleteUser
       })
       .catch((error: any) => {
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -413,7 +413,11 @@ const Users = {
         userId,
       })
       .then(async (result: any) => {
-        await events.sendEvent('', 'b2b-organizations-graphql.removeUser', { id, email })
+        await events.sendEvent('', 'b2b-organizations-graphql.removeUser', {
+          id,
+          email,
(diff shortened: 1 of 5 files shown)


node/resolvers/Mutations/Users.ts
```diff
@@ -412,7 +412,8 @@ const Users = {
         id,
         userId,
       })
-      .then((result: any) => {
+      .then(async (result: any) => {
+        await new Promise(events.sendEvent('', 'b2b-organizations-graphql.removeUser', { id, email }))
```
polishq, line 416: To simplify, you should be able to do this:
```suggestion
        await events.sendEvent('', 'b2b-organizations-graphql.removeUser', { id, email })
```

polishq (changes requested):

node/resolvers/Mutations/Users.ts
rafael-marques: I did this before but for some reason it was not working until I put in a promise like this. I can try once more and I'll give a feedback

polishq: There's also a lint error on this line, but making the above change will likely fix the lint error as well.

rafael-marques: @polishq I changed to the suggested solution and this time it worked. I've just pushed a new commit with the fix.

enzomerca (approved):
LGTM, just wondering if we should add such events to other actions as well (e.g. deleteCostCenter).

Just out of curiosity, what does your script do when an user is deleted @rafael-marques ?

rafael-marques:
Hey @enzomerca thanks for that. I believe applying that to other actions would be helpful.
Our script needs to delete the same user in a CRM we integrated with in order to sync the records, that's why we send the email in the payload.

node/resolvers/Mutations/Users.ts
```diff
@@ -412,7 +412,12 @@ const Users = {
         id,
         userId,
       })
-      .then((result: any) => {
+      .then(async (result: any) => {
+        await events.sendEvent('', 'b2b-organizations-graphql.removeUser', {
```
mairatma, line 416: I'm not very familiar with colossus events, but in general we usually don't wait for events like this to successfully finish before returning the response of an API. We don't want to response to the removeUser API to fail just because we failed to send the event, right? 

So I believe that we should remove the `await` here and let the event be silently sent on the background, returning the response regardless of its success, since the event doesn't alter the actual operation's success.

So what I'm suggesting here is:

```suggestion
       // Send an event on the background, but don't wait to respond.
        events.sendEvent('', 'b2b-organizations-graphql.removeUser', {
```

mairatma:
> Hey @enzomerca thanks for that. I believe applying that to other actions would be helpful. Our script needs to delete the same user in a CRM we integrated with in order to sync the records, that's why we send the email in the payload.

I agree that it can be useful, but let's avoid adding more events unless necessary, ok? Otherwise we'll later be in trouble if we ever need to remove the events, afraid that someone might be using them.

node/resolvers/Mutations/Users.ts
rafael-marques: That makes total sense @mairatma . I've just removed it

mairatma (dismissed):
Thanks, lgtm! Can you just rebase to fix conflicts with the `master` branch?

rafael-marques:
@mairatma looks like it's already up to date with master. Where are the conflicts? 
image

rafael-marques:
Nvm, I've found it in the CHANGELOG and fixed it. Can you review again?

mairatma (approved):

mairatma:
@polishq I always get this kind of problem on external PRs, where one of the checks fails complaining of the missing `safe to test` label. I add it, but the other check removes it before the failing one can verify that it was added... Do you know how to go around this?

polishq:
> @polishq I always get this kind of problem on external PRs, where one of the checks fails complaining of the missing `safe to test` label. I add it, but the other check removes it before the failing one can verify that it was added... Do you know how to go around this?

@mairatma I believe it's behaving as intended. Adding the label triggers the actions to switch to the "pull_request_target" workflow (specifically for external contributions) instead of the normal "pull_request" workflow, and the label is automatically removed as part of that flow. That first security check is intended to remain "Failed" in this scenario, but the PR is still mergeable since it's not a required check.

polishq (approved):
LGTM

albertm805 (approved):
LGTM

PR #127, opened 2023-09-19
feat: add metric on updating organization data

Rudge:

We need to measure the B2B Suite product better to understand the behavior of merchants and buyer organizations. As described in [this doc](https://docs.google.com/document/d/1PdtmfL4rIKpzKUBR6ZGqqixMsfFzuqI38xmtSJ8wki8/edit).

#### How to test it?

- Access some organization in the workspace, path `admin/b2b-organizations/organizations`
- Change some data
- Use the query editor on Redshift to find the event persisted 1 hour after the changes
- Query
```SQL
select 
json_extract_path_text(json_serialize(payload), 'kind') as kind,
json_extract_path_text(json_serialize(payload), 'account') as account,
payload,
ingestion_time
FROM
    vtex.schemaless.b2b_suite_buyerorg_data_raw b2b_raw
where account = 'b2bsuite'
and kind = 'update-organization-graphql-event'
```

[Workspace](https://b2bteam1260--b2bsuite.myvtex.com/admin)

image

I tried to consider just the fields sent from the body, but the request had all fields from the organization page, so it was necessary to compare the data between the organization persisted and the data sent in the request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Add metrics on updating organization data
+
 ## [0.37.0] - 2023-09-19
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-### Added
-
-- Add metrics on updating organization data
-
 ## [0.37.0] - 2023-09-19
 
 
--- a/node/resolvers/Mutations/Organizations.ts
+++ b/node/resolvers/Mutations/Organizations.ts
@@ -18,6 +18,7 @@ import B2BSettings from '../Queries/Settings'
 import CostCenters from './CostCenters'
 import MarketingTags from './MarketingTags'
 import checkConfig from '../config'
+import { sendUpdateOrganizationMetric } from '../../utils/metrics/organization'
 
 const createUserAndAttachToOrganization = async ({
   storefrontPermissions,
@@ -372,15 +373,17 @@ const Organizations = {
       ctx
     )) as B2BSettingsInput
 
+    let currentOrganizationData: Organization = {} as Organization
+
     try {
-      const currentData: Organization = await masterdata.getDocument({
+      currentOrganizationData = await masterdata.getDocument({
         dataEntity: ORGANIZATION_DATA_ENTITY,
         fields: ORGANIZATION_FIELDS,
         id,
       })
 
       if (
-        currentData.status !== status &&
+        currentOrganizationData.status !== status &&
         notifyUsers &&
         settings?.transactionEmailSettings?.organizationStatusChanged
       ) {
@@ -398,21 +401,25 @@ const Organizations = {
     }
 
     try {
+      const fields = {
+        collections,
+        ...((tradeName || tradeName === '') && { tradeName }),
+        customFields,
+        name,
+        paymentTerms,
+        priceTables,
+        ...(salesChannel && { salesChannel }),
+        ...(sellers && { sellers }),
+        status,
+      }
+
       await masterdata.updatePartialDocument({
         dataEntity: ORGANIZATION_DATA_ENTITY,
-        fields: {
-          collections,
-          ...((tradeName || tradeName === '') && { tradeName }),
-          customFields,
-          name,
-          paymentTerms,
-          priceTables,
-          ...(salesChannel && { salesChannel }),
-          ...(sellers && { sellers }),
-          status,
-        },
+        fields,
         id,
       })
+
+      sendUpdateOrganizationMetric(logger, fields, currentOrganizationData)
 
       return { status: 'success', message: '' }
     } catch (error) {
--- a/node/resolvers/Mutations/Organizations.ts
+++ b/node/resolvers/Mutations/Organizations.ts
@@ -373,7 +373,7 @@ const Organizations = {
       ctx
     )) as B2BSettingsInput
 
-    let currentOrganizationData: Organization = {} as Organization
+    let currentOrganizationData: Organization | undefined
 
     try {
       currentOrganizationData = await masterdata.getDocument({
@@ -383,7 +383,7 @@ const Organizations = {
       })
 
       if (
-        currentOrganizationData.status !== status &&
+        currentOrganizationData?.status !== status &&
         notifyUsers &&
         settings?.transactionEmailSettings?.organizationStatusChanged
       ) {
--- a/node/resolvers/Mutations/Organizations.ts
+++ b/node/resolvers/Mutations/Organizations.ts
@@ -419,7 +419,11 @@ const Organizations = {
         id,
       })
 
-      sendUpdateOrganizationMetric(logger, fields, currentOrganizationData)
+      sendUpdateOrganizationMetric(logger, {
+        account: ctx.vtex.account,
+        currentOrganizationData,
+        updatedProperties: fields,
+      })
 
       return { status: 'success', message: '' }
     } catch (error) {
--- a/node/utils/metrics/organization.ts
+++ b/node/utils/metrics/organization.ts
@@ -42,23 +42,31 @@ const getFieldsNamesByFieldsUpdated = (
   const { currentOrganizationData, updatedProperties } =
     updateOrganizationParams
 
-  for (const key in updatedProperties) {
-    if (Object.prototype.hasOwnProperty.call(updatedProperties, key)) {
-      const value = updatedProperties[key as keyof typeof updatedProperties]
+  if (!currentOrganizationData) {
+    return updatedPropName
+  }
 
-      // I tried to compare the objects value !== currentOrganizationData[key as keyof Organization,
-      // but it was not working, so I use JSON.stringify
-      if (
-        JSON.stringify(value) !==
-        JSON.stringify(
-          !currentOrganizationData ||
-            currentOrganizationData[key as keyof Organization]
-        )
+  Object.entries(updatedProperties).forEach(
+    ([updatedPropertyKey, updatedPropertyValue]) => {
+      if (updatedPropertyValue instanceof Array) {
+        // I tried to compare the objects value !== currentOrganizationData[key as keyof Organization,
+        // but it was not working, so I use JSON.stringify
+        if (
+          JSON.stringify(updatedPropertyValue) !==
+          JSON.stringify(
+            currentOrganizationData[updatedPropertyKey as keyof Organization]
+          )
+        ) {
+          updatedPropName.push(updatedPropertyKey)
+        }
+      } else if (
+        updatedPropertyValue !==
+        currentOrganizationData[updatedPropertyKey as keyof Organization]
       ) {
-        updatedPropName.push(key)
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,9 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
 
-## [0.37.0] - 2023-09-19
```
mairatma, line 10: I believe you removed this line by mistake, can you move it to after the new line you've added below? It should be before line 13.

node/utils/metrics/organization.ts
```diff
@@ -0,0 +1,96 @@
+import type { Logger } from '@vtex/api/lib/service/logger/logger'
+
+import type { Metric } from './metrics'
+import { sendMetric } from './metrics'
+
+interface UpdateOrganizationFieldsMetric {
+  update_details: { properties: string[] }
```
mairatma, line 7: Why use underscore_case here instead of camelCase like the rest? We usually only do this when we have no choice (e.g. a backend request that requires underscore_case on the body fields).

Does Redshift data usually use underscore? If so that's fine, just wanted to be sure.

node/utils/metrics/organization.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  for (const key in updatedProperties) {
+    if (Object.prototype.hasOwnProperty.call(updatedProperties, key)) {
+      const value = updatedProperties[key as keyof typeof updatedProperties]
+
+      // I tried to compare the objects value !== currentOrganizationData[key as keyof Organization,
+      // but it was not working, so I use JSON.stringify
+      if (
+        JSON.stringify(value) !==
+        JSON.stringify(currentOrganizationData[key as keyof Organization])
```
mairatma, line 53: This kind of comparison is kind of expensive, depending on the size of the objects being stringified. It's also not guaranteed to work, as stringified objects don't have any guarantees regarding the order of keys, for example. So the objects might be the same, but the stringified versions might not match due to sorting.

Do we really need to do this level of comparison? If it's really essential, it might be better to analyze the fields that can be updated and see how complex they are, to see if we can do simpler comparisons for different types of fields.

node/utils/metrics/organization.ts
```diff
@@ -0,0 +1,96 @@
...
+const getFieldsNamesByFieldsUpdated = (
+  updateOrganizationParams: UpdateOrganizationParams
+): string[] => {
+  const updatedPropName: string[] = []
+
+  const { currentOrganizationData, updatedProperties } =
+    updateOrganizationParams
+
+  for (const key in updatedProperties) {
+    if (Object.prototype.hasOwnProperty.call(updatedProperties, key)) {
```
mairatma, line 46: Since we're already looping through the keys within `updatedProperties`, this call shouldn't be necessary.

Also, when it is necessary, you should be able to call the function directly on the object, like this:
```suggestion
    if (updatedProperties.hasOwnProperty(key)) {
```

node/utils/metrics/organization.ts
```diff
@@ -0,0 +1,96 @@
...
+        updatedPropName.push(key)
+      }
+    }
+  }
+
+  return updatedPropName
+}
+
+export interface UpdatedOrganizationProps {
+  priceTables: any[]
```
mairatma, line 64: In Typescript we always avoid using `any`. With the right tsconfig this should be throwing lint warnings or even errors actually.

For use cases where the actual detailed type can't be known in advance, or we don't have the definitions anywhere, you can instead use `unknown`, which is stricter than `any` (it doesn't allow accessing fields of the variable, for example, while `any` will allow anything).

In this case though, I believe that you can be even more generic. The actual keys here don't matter much to you, you just need to know that this will be a map of string to unknown values, so this type could just be:

```js
export type UpdatedOrganizationProps = Record<string, unknown>
```

This should also help with the code that was trying to access keys within this object, within `getFieldsNamesByFieldsUpdated`.

node/resolvers/Mutations/Organizations.ts
```diff
@@ -372,15 +373,17 @@ const Organizations = {
       ctx
     )) as B2BSettingsInput
 
+    let currentOrganizationData: Organization = {} as Organization
```
mairatma, line 376: From the code here in this file I can see that  `currentOrganizationData` may end up being just `{}` instead of a real `Organization` object. `{}` is not a valid `Organization` type though, and that's probably why you had to add the casting for `as Organization`.

It's safer to be clearer here that this variable may not have a value, like this:

```suggestion
    let currentOrganizationData: Organization  | undefined
```

Then you can deal directly with this use case within `sendUpdateOrganizationMetric`, either by setting its default value to `{}` (and so the type there should be `Organization | {}` ) or by skipping the entire process in that case.

node/utils/metrics/organization.ts
mairatma: Or even better, you can just use the `Organization` type directly, shouldn't they be basically the same? The update object is probably the same type, but with all keys being optional. In Typescript you can achieve that like this:

```js
export type UpdateOrganizationProps = Partial<Organization>
```

That will also help when comparing their values, as the keys will now match, and you won't need to cast types around.

mairatma (commented):
Nice to see this going forward! 👏

CHANGELOG.md
Rudge: I don't know why, but the previous build didn't work and blocked me. Do you know how to solve it?
https://github.com/vtex-apps/b2b-organizations-graphql/actions/runs/6241791489
https://github.com/vtex-apps/b2b-organizations-graphql/actions/runs/6248666170
https://github.com/vtex-apps/b2b-organizations-graphql/actions/runs/6248730389
https://github.com/vtex-apps/b2b-organizations-graphql/actions/runs/6248791700
https://github.com/vtex-apps/b2b-organizations-graphql/actions/runs/6248760613

node/utils/metrics/organization.ts
Rudge: I found [this definition](https://internal-docs.vtex.com/Analytics/VTEX-Data-Platform/data-ingestion-API/?h=schemaless#schema-enforced-ingestion-adding-new-data-sources) for Redshift and previous [Mauro used snake_case too](https://github.com/vtex-apps/b2b-organizations-graphql/blob/master/node/utils/metrics/impersonate.ts#L6).

node/utils/metrics/organization.ts
Rudge: Yes, we need to make a diff between the persisted info and the changed. Do you have a suggestion to simplify?

CHANGELOG.md
Rudge: I did a rebase, and now it is worked.

node/utils/metrics/organization.ts
Rudge: I tried to use it, but the lint suggested the change. Can I keep this eslint error?
image

node/resolvers/Mutations/Organizations.ts
Rudge: It is OK for me, but with this approach, we need to cover more branches cases in the tests.

node/utils/metrics/organization.ts
mairatma: Got it, so I think it's suggesting the use of `Object.hasOwnProperty(updatedProperties, key)`. I just wouldn't use `Object.prototype` here, and also not use `.call(`, for example.

node/utils/metrics/organization.ts
mairatma: It depends on the contents of that object. The more nested it is, the more complex it will be. Do we know what keys may be on the object and what they're contents are? If you can grab a good example, as complete as possible, we can analyse it better to see the best way to do the comparison.

node/utils/metrics/organization.ts
mairatma: But I still think this line shouldn't be necessary. Fixing the type of the updated fields to be based on Organization will probably remove any lint requesting this check to be made.

node/resolvers/Mutations/Organizations.ts
mairatma: @Rudge that's ok. I think we shouldn't make coverage affect the mergeability of our PRs for B2B Suite though. The tests are not very trustworthy yet, the coverage doesn't say much about their quality or real use case coverage today. 

It's better to use the types correctly, avoiding future bugs. The types are a stricter test than the tests themselves in this case. It won't allow the code to skip dealing correctly with the edge case, and that's better than covering the use case with a test that may not even do the right checks to be counted towards code coverage.

node/utils/metrics/organization.ts
Rudge: The example is the organization structure
```Typescript
interface Organization {
  id: string
  name: string
  tradeName?: string
  collections: Collection[]
  costCenters: string[]
  paymentTerms: PaymentTerm[]
  priceTables?: Price[]
  status: string
  created: string
  customFields?: CustomField[]
  salesChannel?: string
  sellers?: Seller[]
}
```

Rudge: What do you think to use [`isEqual` from `lodash`](https://www.npmjs.com/package/@types/lodash.isequal)?

mairatma: @Rudge isEqual would be more trustworthy than the JSON.stringify, as it won't fail due to sorting of keys.

It will still be expensive though, so I'd prefer avoiding if possible. At least this happens on the background though, so we won't delay the response by doing it, just spend more compute resources.

Do you have real examples of an update object and the original organization object? The types are not very trustworthy, they may be missing fields or being simpler than the real thing. With a real example this can be more tangible, so that we have a better idea of the sizes as well as the type of comparisons that will be made.

Rudge: A real example is it.
```JSON
{
    "id": "3956085d-36b8-11ee-83ab-0e170ba3b9a5",
    "name": "BAHIA EVENTOS E LOCAÇÕES LTDA.",
    "tradeName": "BA EVENTOS",
    "status": "active",
    "created": "2023-08-09T13:25:36.387Z",
    "collections": [],
    "paymentTerms": [
        {
            "id": "6",
            "name": "Boleto Bancário",
        },
        {
            "id": "999999",
            "name": "Credit card",
        },
        {
            "id": "125",
            "name": "Pix"
        },
        {
            "id": "172",
            "name": "Pagaleve"
        }
    ],
    "priceTables": [
        "pjba"
    ],
    "salesChannel": null,
    "costCenters": [],
    "sellers": null,
    "customFields": []
}
```

mairatma: Esse aí é qual dos dois objetos? E pode colocar o outro também (se esse for a org original, colar o obj de update, ou vice-versa). Peço porque muitas vezes esses updates são bem mais simples do que a estrutura completa do objeto.

mairatma: Mas pelo que tou vendo me parece um objeto bem simples. Com isEqual do lodash ficaria uma comparação mais garantida, e por ser um objeto simples não vai ficar tão caro. Rodando em background como está é melhor ainda. Me parece ok, fico mais tranquila assim!

Rudge: Esse é o objeto que é enviado para atualizar e contem a mesma estrutura do objeto original. Lembrando que toda vez que muda um campo na organização a estrutura é enviada inteiramente e cada estrutura dessa é uma aba ou uma seção dessa tela.
image

enzomerca (approved):
LGTM

PR #130, opened 2023-10-04
fix: impersonate user with sales role

Rudge:

The users with the `sales` roles can't impersonate users, even if they have permission.

#### How to test it?

- Open the path /account#/organization, log with sales role account
- Go to Users
- Click in the 3 points to impersonate the user.

[Workspace](https://www.rivieraworks.ro/account?workspace=ki910328#/organization)

Screen Recording 2023-10-05 at 17 27 10

### DIFF
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -48,7 +48,7 @@ const isUserPermitted = ({
   roleSlug,
 }: any) =>
   (storefrontPermissions?.permissions?.includes('add-users-organization') &&
-    sessionData.namespaces['storefront-permissions'].organization.value ===
+    sessionData.namespaces['storefront-permissions']?.organization?.value ===
       orgId &&
     !roleSlug.includes('sales')) ||
   (storefrontPermissions?.permissions?.includes('add-sales-users-all') &&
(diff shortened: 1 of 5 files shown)


node/resolvers/Mutations/Users.ts
```diff
@@ -183,26 +183,29 @@ const Users = {
...
-      let permitted = false
+      const canImpersonateUsersWithCostCenterPermission =
+        !roleSlug.includes('admin') &&
+        storefrontPermissions?.permissions?.includes(
+          'impersonate-users-costcenter'
+        ) &&
+        user?.costId ===
+          sessionData?.namespaces['storefront-permissions']?.costcenter?.value
 
-      if (!roleSlug.includes('sales')) {
```
enzomerca, line 188: It looks like the code was not allowing `sales` role to impersonate on purpose. Do we have any idea why?

node/resolvers/Mutations/Users.ts
```diff
@@ -183,26 +183,29 @@ const Users = {
       const result = await storefrontPermissionsClient.getRole(user.roleId)
       const roleSlug = result?.data?.getRole?.slug
 
-      let permitted = false
+      const canImpersonateUsersWithCostCenterPermission =
+        !roleSlug.includes('admin') &&
```
enzomerca, line 187: I also wonder why we need to check if role is not `admin`.  Admins can't impersonate "with cost centers"? But can impersonate on the other cases?

node/resolvers/Mutations/Users.ts
Rudge: I don't know why, but in the [documentation](https://developers.vtex.com/docs/apps/vtex.b2b-organizations) it is specified that the role has access. Do you know the reasons @wender or @arturmagalhaesjr?
Screenshot 2023-10-06 at 10 37 34

node/resolvers/Mutations/Users.ts
Rudge: I kept the implementation, but now, thinking more, let's remove all roles and just use permissions.

node/resolvers/Mutations/Users.ts
```diff
@@ -300,13 +299,14 @@ const Users = {
           ) &&
             !roleSlug.includes('admin') &&
```
enzomerca, line 300: Are we keeping the "admin" role check here? Basically we are saying that, if this user has the `impersonate-users-costcenter` permission but is an `admin`, it can't impersonate, which doesn't make sense I think...

Rudge: I forgot. Thanks. I changed `impersonateB2BUser`, but I will change `impersonateUser` too.

enzomerca (approved):

enzomerca (approved):

PR #134, opened 2023-10-30
feat: add audit metrics for authenticated access

Rudge:

We need to monitor the access in the app, checking the graphql operations with authentication access.

#### How to test it?

- Access some audited operation from graphql.
getOrganizationById
getCostCenterById
getOrganizationByIdStorefront
getCostCenterByIdStorefront
getSalesChannels
getCostCentersByOrganizationId
getOrganizations
getOrganizationsByEmail
getB2BSettings
getMarketingTags
createOrganizationRequest
getBinding
getPaymentTerms

[Workspace](https://securitymetrics--b2bsuite.myvtex.com/)

image

### DIFF
--- a/node/resolvers/directives/auditAccess.ts
+++ b/node/resolvers/directives/auditAccess.ts
@@ -21,10 +21,7 @@ export class AuditAccess extends SchemaDirectiveVisitor {
     }
   }
 
-  private async sendAuthMetric(
-    field: GraphQLField<any, any>,
-    context: Context
-  ) {
+  private async sendAuthMetric(field: GraphQLField<any, any>, context: any) {
     const {
       clients: { storefrontPermissions },
       vtex: { adminUserAuthToken, storeUserAuthToken, account, logger },
@@ -33,7 +30,11 @@ export class AuditAccess extends SchemaDirectiveVisitor {
 
     const operation = field.astNode?.name?.value ?? request.url
     const forwardedHost = request.headers['x-forwarded-host'] as string
-    const caller = request.headers['x-vtex-caller'] as string
+    const caller =
+      context?.graphql?.query?.senderApp ??
+      context?.graphql?.query?.extensions?.persistedQuery?.sender ??
+      request.header['x-b2b-senderapp'] ??
+      (request.headers['x-vtex-caller'] as string)
 
     const hasAdminToken = !!(
       adminUserAuthToken ?? (context?.headers.vtexidclientautcookie as string)
--- a/node/resolvers/directives/auditAccess.ts
+++ b/node/resolvers/directives/auditAccess.ts
@@ -15,49 +15,54 @@ export class AuditAccess extends SchemaDirectiveVisitor {
       context: Context,
       info: any
     ) => {
-      const {
-        clients: { storefrontPermissions },
-        vtex: { adminUserAuthToken, storeUserAuthToken, account, logger },
-        request,
-      } = context
-
-      const operation = field.astNode?.name?.value ?? request.url
-      const forwardedHost = request.headers['x-forwarded-host'] as string
-      const caller = request.headers['x-vtex-caller'] as string
-
-      const hasAdminToken = !!(
-        adminUserAuthToken ?? (context?.headers.vtexidclientautcookie as string)
-      )
-
-      const hasStoreToken = !!storeUserAuthToken
-      const hasApiToken = !!request.headers['vtex-api-apptoken']
-
-      let role
-      let permissions = []
-
-      if (hasAdminToken || hasStoreToken) {
-        const userPermissions = await getUserPermission(storefrontPermissions)
-
-        if (userPermissions) {
-          role = userPermissions.role?.slug
-          permissions = userPermissions.permissions
-        }
-      }
-
-      const authMetric = new AuthMetric(account, {
-        caller,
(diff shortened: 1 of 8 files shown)


node/resolvers/directives/auditAccess.ts
```diff
@@ -0,0 +1,63 @@
...
+      )
+
+      const hasStoreToken = !!storeUserAuthToken
+      const hasApiToken = !!request.headers['vtex-api-apptoken']
+
+      let role
+      let permissions = []
+
+      if (hasAdminToken || hasStoreToken) {
+        const checkUserPermissions = await getCheckUserPermission(
```
mairatma, line 39: Same here as I commented for the [other similar PR](https://github.com/vtex-apps/storefront-permissions/pull/118#discussion_r1377481834)

node/resolvers/directives/auditAccess.ts
```diff
@@ -0,0 +1,63 @@
...
+          storefrontPermissions
+        )
+
+        role = checkUserPermissions.role.slug
```
mairatma, line 43: I checked the `getCheckUserPermission` function further below and it seems like it can return `null`, so this code should be able to handle that.

node/resolvers/directives/withPermissions.ts
```diff
@@ -4,6 +4,15 @@ import type { GraphQLField } from 'graphql'
 import { defaultFieldResolver } from 'graphql'
 import { SchemaDirectiveVisitor } from 'graphql-tools'
 
+import type StorefrontPermissions from '../../clients/storefrontPermissions'
+
+export const getCheckUserPermission = async (
```
enzomerca, line 9: We are just getting the user permissions here right? Should we call it just  `getUserPermission`?

node/resolvers/directives/withPermissions.ts
```diff
@@ -4,6 +4,15 @@ import type { GraphQLField } from 'graphql'
...
+  storefrontPermissions: StorefrontPermissions
+) =>
+  storefrontPermissions.checkUserPermission().then((result: any) => {
+    return result?.data?.checkUserPermission ?? null
+  })
```
mairatma, line 14: At VTEX we prefer using async/await instead of promises (meaning `.then()`, `.catch()`, etc). In the end they're the same thing, as async/await uses promises behind the scenes, but async/await has better readability, so that's why it's preferred (outside VTEX as well).

You can make any function using promises use async/await instead, by replacing `.then` with `await` and `.catch` with a regular try/catch block. This is what it would look like here:

```suggestion
export const getCheckUserPermission = async (
  storefrontPermissions: StorefrontPermissions
) =>
  const result = await storefrontPermissions.checkUserPermission()
  return result?.data?.checkUserPermission ?? null
```

Behind the scenes, the function above will be compiled into a very similar version to what you have in this PR (using promises). It's sugar syntax, same behavior.

node/utils/metrics/auth.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface AuthAuditMetric {
+  operation: string
+  forwardedHost: string
+  caller: string
+  role: string
+  permissions: string[]
+  hasAdminToken: boolean
+  hasStoreToken: boolean
+  hasApiToken: boolean
```
enzomerca, line 14: Do we care which of these tokens its being used? Maybe we should emit just a `hasToken` field aggregating all of these to reduce the amount of data we emit?

node/resolvers/directives/withPermissions.ts
```diff
@@ -21,21 +30,18 @@ export class WithPermissions extends SchemaDirectiveVisitor {
...
+      ).catch((error: any) => {
+        if (!adminUserAuthToken) {
+          logger.error({
+            message: 'getPermissionsError',
+            error,
+          })
+        }
+
+        return null
+      })
```
mairatma, line 44: I won't ask you to change this one, as it was already here before. But just to further explain, this code here is equivalent to:

```suggestion
      try {
        appClients.storefrontPermissions = await getCheckUserPermission(
          storefrontPermissions
        )
      } catch (error) {
        if (!adminUserAuthToken) {
          logger.error({
            message: 'getPermissionsError',
            error,
          })
        }

        // The return value from `.catch()` promise functions is the value returned by the promise,
        // so it'd be set to `appClients.storefrontPermissions` in the original code.
        appClients.storefrontPermissions = null
      }
```

node/resolvers/directives/withPermissions.ts
Rudge: I agree, I let the same previous name be created. But I will change.

node/utils/metrics/auth.ts
Rudge: The intention was to monitor which token was used for each operation. I didn't use just one because I'm not sure if we should allow multiple tokens in some cases.

node/resolvers/directives/auditAccess.ts
mairatma: @Rudge this one is the only one still pending an update, to avoid awaiting within the resolver function.

Rudge: Sorry, could you look again, please?

mairatma (dismissed):

enzomerca (approved):

PR #135, opened 2023-11-03
Fix the costCenterId

rbussola:
Fix the costCenterId when using mutation createOrganization

When using the createOrganization mutation, there is no costCenterId returned in the response:

image

#### How to test it?

- Access the graphQL IDE in [internal account](https://tkt928918--b2bsuite.myvtex.com/admin/graphql-ide)
- Create a organization using createOrganization mutation

- Before
image

-After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


mairatma (approved):

mairatma:
Just curious, did this never work before or was it a regression?

rbussola:
@mairatma it was a regression introduced when we released version 0.42.0

PR #139, opened 2023-11-17
feat: add createUserWithEmail graphql API

enzomerca:

Add a `createUserWithEmail` graphql with @checkAdminAccess directive to be called by the b2b-bulk-import app. This `createUserWithEmail` has stricter params than `addUser` and ultimately calls the storefront permissions addUser (just as the `addUser` graphql api).
This is also to avoid touching on the permissions of `addUser` that uses the @checkUserAccess directive.

#### How to test it?

Try to create an user calling the `createUserWithEmail` API providing the expected parameters (note only `canImpersonate` is optional):
```
  createUser(
    orgId: ID!
    costId: ID!
    roleId: ID!
    name: String!
    email: String!
    canImpersonate: Boolean = false
  )
```

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -203,6 +203,16 @@ type Mutation {
     @withPermissions
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
+  createUser(
+    orgId: ID!
+    costId: ID!
+    roleId: ID!
+    name: String!
+    email: String!
+    canImpersonate: Boolean = false
+  ): MutationResponse
+    @checkAdminAccess
+    @cacheControl(scope: PRIVATE)
   updateUser(
     id: ID
     roleId: ID!
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -188,6 +188,13 @@ type Mutation {
     @withPermissions
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
+  """
+    addUser will create an user with the provided parameters.
+    This is called either from the Admin UI or the storefront UI (where
+    the buyer org manager/admin can add a new member to its own org/cost center).
+    Although the mutation allows an id/userId/clId to be provided, these are not
+    exposed/collected on the UIs.
+  """
   addUser(
     id: ID
     roleId: ID!
@@ -203,7 +210,18 @@ type Mutation {
     @withPermissions
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
-  createUser(
+  """
+    createUserWithEmail is a simplified version of addUser. Both
+    mutations will create an user for the org/cost center,
+    but createUserWithEmail expect all fields (except canImpersonate) as
+    required and does not allow the user to provide its own id/userId/clId,
+    which will be automatically created by the mutation. In this case, the
+    email is clearly used as an identifier for the user. This is currently
+    used by the bulk import use case, but could be used by other use cases
+    as well. This function also has stricter permissions to be used by the
+    store admin only, but could have less strict permissions in the future.
+  """
+  createUserWithEmail(
     orgId: ID!
     costId: ID!
     roleId: ID!
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -520,12 +520,16 @@ const Users = {
     } = ctx as any
 
     try {
-      const result = await storefrontPermissionsClient
-        .addUser({ orgId, costId, roleId, name, email, canImpersonate })
-
+      const result = await storefrontPermissionsClient.addUser({
+        orgId,
+        costId,
+        roleId,
+        name,
+        email,
+        canImpersonate,
+      })
       return result.data.addUser
-
-    } catch (error: any) {
+    } catch (error) {
       logger.error({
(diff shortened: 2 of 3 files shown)


enzomerca:
I will hold merging this PR until after black friday 2023.

graphql/schema.graphql
```diff
@@ -203,6 +203,16 @@ type Mutation {
     @withPermissions
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
+  createUser(
```
mairatma, line 206: I understand the idea of creating a new mutation, but I worry because it'll be very confusing to have both `addUser` and `createUser`. The names don't make it clear when to use one or the other, or any differences in behavior. 

The main difference seems to be the enforced admin access. Why exactly does this `createUser` need this special permission check? Is the behavior very different, in a way that we should restrict this use case to admin users?

If so, maybe call this one `createUserFromAdmin`, so that this is clearer for the caller. It would also be helpful to add a comment (or comments, one per mutation) explaining the main differences between them.

node/resolvers/Mutations/Users.ts
```diff
@@ -509,6 +509,44 @@ const Users = {
...
+        } else if (
+          message.includes(MessageSFPUserAddError.DUPLICATED_ORGANIZATION)
+        ) {
+          status = StatusAddUserError.DUPLICATED_ORGANIZATION
+        } else {
+          status = StatusAddUserError.ERROR
+        }
+
+        return { status, message }
+      })
```
mairatma, line 547: I know that this repo sometimes uses promises and sometimes uses async/await, but in 99.99% of cases we avoid the Promise syntax (async/await is more readable and so it's the usual pattern in VTEX and outside it as well). Could you switch to async/await syntax? It's always possible, since it's just a syntax thing. It would be something like this:

```suggestion
    try {
      const result = await storefrontPermissionsClient
        .addUser({ orgId, costId, roleId, name, email, canImpersonate })
      return result.data.addUser
    } catch (error: any) {
      logger.error({
        error,
        message: 'addUser-error',
      })

      const message = error.graphQLErrors[0]?.message ?? error.message
      let status = ''

      if (message.includes(MessageSFPUserAddError.DUPLICATED)) {
        status = StatusAddUserError.DUPLICATED
      } else if (
        message.includes(MessageSFPUserAddError.DUPLICATED_ORGANIZATION)
      ) {
        status = StatusAddUserError.DUPLICATED_ORGANIZATION
      } else {
        status = StatusAddUserError.ERROR
      }

      return { status, message }
    }
```

graphql/schema.graphql
enzomerca: We recently added the following mutations for the bulk-import use case:
```
  createOrganizationAndCostCentersWithId(
    input: OrganizationInput!
  ): MasterDataResponse
    @checkAdminAccess
    @cacheControl(scope: PRIVATE)

  createCostCenterWithId(
    organizationId: ID
    input: CostCenterInput!
  ): MasterDataResponse
    @checkAdminAccess
    @cacheControl(scope: PRIVATE)
```

As you can see, both have the same permission check. As we are adding the new `createUser` for the same bulk-import use case, I wanted to keep the same permission check as these other 2 mutations. I didn't want to change the permissions for the `addUser` mutation to avoid breaking it.
On top of that, this `createUser` mutation is cleaner than the other as it expects less parameters, explicitly expects the required fields (using `ID!`) and does only the necessary permission checks (while `addUser` does additional user/roles permission checks before creating the user).

I agree it is not great to have 2 mutations with very similar names like this, so let me think of a better name for it. I can also add a comment on top of each function just to clarify when to use each one.

node/resolvers/Mutations/Users.ts
enzomerca: Sounds good! I just copied the code from `addUser`, but in this case I see no harm in changing it, as one of the ideas is to have a "revamped"/better `addUser` function

node/resolvers/Mutations/Users.ts
```diff
@@ -509,6 +509,46 @@ const Users = {
...
+        email,
+        canImpersonate,
+      })
+      return result.data.addUser
+    } catch (error) {
+      logger.error({
+        error,
+        message: 'addUser-error',
+      })
+      const message = error.graphQLErrors[0]?.message ?? error.message
```
mairatma, line 537: This `error.graphQLErrors` object may not be defined depending on the error, so you also need one more `?.` operator here:

```suggestion
      const message = error.graphQLErrors?.[0]?.message ?? error.message
```

node/resolvers/Mutations/Users.ts
```diff
@@ -509,6 +509,46 @@ const Users = {
...
+        roleId,
+        name,
+        email,
+        canImpersonate,
+      })
+      return result.data.addUser
+    } catch (error) {
+      logger.error({
+        error,
+        message: 'addUser-error',
```
mairatma, line 535: This is the same message for the other `addUser` function, maybe better to have a different one here so we can distinguish them better, right?

mairatma (dismissed):
Lgtm, just 2 suggestions inline.

mairatma (approved):

PR #140, opened 2023-11-29
feat: add token validation and token to call storefront-permission

Rudge:

We found some apps that don't pass the token to call graphql queries or mutations. So we need a guarantee to have all integrations send the token, and we will block unauthenticated requests from the app provider.

#### How to test it?

- Open the [page organizations](https://b2bteam1285--b2bstore005.myvtex.com/admin/b2b-organizations/organizations#/organizations) in the admin. `test getOrganizations`
- Select [one organization](https://b2bteam1285--b2bstore005.myvtex.com/admin/b2b-organizations/organizations/d9e51195-0c39-11ed-835d-0ac601682b29/#/default) from page organizations. `test getOrganizationById`
- Open the [page My quotes](https://b2bteam1285--b2bstore005.myvtex.com/b2b-quotes). `test getOrganizationByIdStorefront and getOrganizationsByEmail`
- In the page My quotes, filter the quotes by cost center. `test getCostCentersByOrganizationId`
- Open the page organizations in the admin, select an organization and open the [tab payment terms](https://b2bteam1285--b2bstore005.myvtex.com/admin/b2b-organizations/organizations/d9e51195-0c39-11ed-835d-0ac601682b29/#/pay-terms). `test getPaymentTerms`
- Open the page organizations in the admin, select an organization and open the [tab sales channel](https://b2bteam1285--b2bstore005.myvtex.com/admin/b2b-organizations/organizations/d9e51195-0c39-11ed-835d-0ac601682b29/#/sales-channel). `test getSalesChannels`

[Workspace](https://b2bteam1285--b2bstore005.myvtex.com/)

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -25,7 +25,8 @@ type Query {
     @cacheControl(scope: PRIVATE)
 
   getOrganizationById(id: ID): Organization
-    @cacheControl(scope: PRIVATE, maxAge: SHORT)
+    @cacheControl(scope: PUBLIC, maxAge: SHORT)
+    @auditAccess
   getOrganizationByIdStorefront(id: ID): Organization
     @withSession
     @cacheControl(scope: PRIVATE)
@@ -56,7 +57,8 @@ type Query {
     sortedBy: String = "name"
   ): CostCenterResult @withSession @cacheControl(scope: PRIVATE)
   getCostCenterById(id: ID!): CostCenter
-    @cacheControl(scope: PRIVATE, maxAge: SHORT)
+    @cacheControl(scope: PUBLIC, maxAge: SHORT)
+    @auditAccess
   getCostCenterByIdStorefront(id: ID): CostCenter
     @withSession
     @cacheControl(scope: PRIVATE, maxAge: SHORT)
@@ -94,8 +96,11 @@ type Query {
     @cacheControl(scope: PRIVATE, maxAge: SHORT)
     @checkUserAccess
   getMarketingTags(costId: ID!): MarketingTags
-    @cacheControl(scope: PRIVATE, maxAge: SHORT)
-  getB2BSettings: B2BSettings @cacheControl(scope: PRIVATE, maxAge: SHORT)
+    @cacheControl(scope: PUBLIC, maxAge: SHORT)
+    @auditAccess
+  getB2BSettings: B2BSettings
+    @cacheControl(scope: PUBLIC, maxAge: SHORT)
+    @auditAccess
   getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
 }
 
@@ -104,7 +109,7 @@ type Mutation {
   createOrganizationRequest(
     input: OrganizationInput!
     notifyUsers: Boolean
-  ): MasterDataResponse @checkUserAccess
+  ): MasterDataResponse
   updateOrganizationRequest(
     id: ID!
     status: String!
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -25,8 +25,8 @@ type Query {
     @cacheControl(scope: PRIVATE)
 
   getOrganizationById(id: ID): Organization
-    @cacheControl(scope: PUBLIC, maxAge: SHORT)
-    @auditAccess
+    @cacheControl(scope: PRIVATE)
+    @checkUserAccess
   getOrganizationByIdStorefront(id: ID): Organization
     @withSession
     @cacheControl(scope: PRIVATE)
@@ -57,8 +57,8 @@ type Query {
     sortedBy: String = "name"
   ): CostCenterResult @withSession @cacheControl(scope: PRIVATE)
   getCostCenterById(id: ID!): CostCenter
-    @cacheControl(scope: PUBLIC, maxAge: SHORT)
-    @auditAccess
+    @cacheControl(scope: PRIVATE, maxAge: SHORT)
+    @checkUserAccess
   getCostCenterByIdStorefront(id: ID): CostCenter
     @withSession
     @cacheControl(scope: PRIVATE, maxAge: SHORT)
@@ -80,8 +80,8 @@ type Query {
     @checkUserAccess
     @cacheControl(scope: PRIVATE, maxAge: SHORT)
   getPaymentTerms: [PaymentTerm]
-    @cacheControl(scope: PRIVATE, maxAge: SHORT)
-    @checkUserAccess
+    @cacheControl(scope: PUBLIC, maxAge: SHORT)
+    @auditAccess
 
   getOrganizationsByEmail(email: String): [B2BOrganization]
     @checkUserAccess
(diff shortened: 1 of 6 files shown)


node/clients/storefrontPermissions.ts
```diff
@@ -22,35 +21,49 @@ export default class StorefrontPermissions extends AppGraphQLClient {
...
-          provider: 'vtex.storefront-permissions@1.x',
-          sender: 'vtex.b2b-organizations@0.x',
+    return this.graphql.query(
+      {
+        extensions: {
+          persistedQuery: {
+            provider: 'vtex.storefront-permissions@1.x',
+            sender: 'vtex.b2b-organizations-graphql@0.x',
+          },
+        },
```
enzomerca, line 57: as this is basically the same for all queries, is there a way to parametrize this? so in the future we can change parameters (e.g. the app version) in a single place if needed?

node/clients/storefrontPermissions.ts
```diff
@@ -168,49 +189,45 @@ export default class StorefrontPermissions extends AppGraphQLClient {
...
-          provider: 'vtex.storefront-permissions@1.x',
-          sender: 'vtex.b2b-organizations@0.x',
-        },
-      },
-      query: getB2BUser,
-      variables: { id },
-    })
-  }
-
-  public checkImpersonation = async (): Promise<any> => {
```
enzomerca, line 201: why did we remove the checkImpersonation from this file?

graphql/schema.graphql
```diff
@@ -76,22 +80,21 @@ type Query {
...
   getOrganizationsByEmail(email: String): [B2BOrganization]
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
-    @auditAccess
 
   checkOrganizationIsActive(id: String): Boolean @cacheControl(scope: PRIVATE)
   getSalesChannels: [Channels]
-    @cacheControl(scope: PUBLIC, maxAge: SHORT)
-    @auditAccess
+    @cacheControl(scope: PRIVATE, maxAge: SHORT)
```
mairatma, line 92: What's the rationale for deciding which of these should use PRIVATE vs PUBLIC cache control? I understand that payment terms (further above in this file) seems like it should be private, depending on what it is. But sales channels and binding doesn't seem sensitive, for example, so having a public cache control seems better, since we'll be able to cache at the CDN level. Same question goes for adding @checkUserAccess for these. Aren't those available in the public VTEX rest apis today already?

I may be wrong though, just want to be sure that we don't require auth and cancel CDN caching unnecessarily.

graphql/schema.graphql
```diff
@@ -106,7 +109,7 @@ type Mutation {
   createOrganizationRequest(
     input: OrganizationInput!
     notifyUsers: Boolean
-  ): MasterDataResponse @auditAccess
+  ): MasterDataResponse
```
mairatma, line 112: Just confirming: for this one we've decided to not require any auth right? It makes sense since this looks like it can be called from the storefront, but just wanted to be sure.

graphql/schema.graphql
```diff
@@ -266,6 +265,7 @@ type Mutation {
     @withSession
     @withPermissions
     @cacheControl(scope: PRIVATE)
+    @checkUserAccess
```
mairatma, line 268: What does the `@withPermissions` directive (two lines above) check? Asking so we can avoid calling store-permissions/vtexid twice unnecessarily.

graphql/schema.graphql
Rudge: Good questions, but I don't know if salesChannels channels are sensitive or not. The scope `PRIVATE` was necessary because of the token, but if the information is not sensitive, we can rollback the config. For B2B, we don't have too many requests, so we need to measure if the cost of CDN is higher or not than requests for app. How does the cache control work? Is it just for the operation name or does it have another parameter in the key? I have a lot of questions to decide.

graphql/schema.graphql
Rudge: Exactly! The page is usage in the storefront.

graphql/schema.graphql
Rudge: It calls the `checkUserPermission` in the storefront-permission, so at the moment we are not reviewing the permission flow, just the authentication.

graphql/schema.graphql
mairatma: I'm almost certain that sales channels are public, one of them is always present in the cookie by the way. So I think it's worth confirming with someone so we can decide if this needs auth or not, my guess is that it doesn't.

> For B2B, we don't have too many requests, so we need to measure if the cost of CDN is higher or not than requests for app.

The requests might be low compared to B2C, but I believe it's still worth making the right decisions here so we don't have trouble in the future. Also, from what I remember store-permissions was causing a lot of load on VBase a while ago, and requiring auth here would call store-permissions, right? Avoiding to use auth when unnecessary is good.

> How does the cache control work? Is it just for the operation name or does it have another parameter in the key?

Cache control for GraphQL APIs work per field. So if a request is made with only this `getSalesChannel` field in the root query, that value will be used as the max age for both internal caches (such as Kube Router) as well as the CDN (as long as it's PUBLIC). If a GraphQL request contains multiple query fields, the minimum max age is used, and it will be considered PRIVATE if at least one of them is configured as such. But from what I saw so far we're always sending requests with a single root query field at a time.

graphql/schema.graphql
mairatma: Got it, the permissions check is the next level, then, doing authorization instead of authentication.

In that case it seems like `checkUserPermission` should always run before `withPermissions` when both exist. It doesn't make sense to check permissions for an unauthenticated or invalidly authenticated user.

node/resolvers/Routes/index.ts
```diff
@@ -38,7 +38,8 @@ const getUserAndPermissions = async (ctx: Context) => {
   const {
     data: { checkUserPermission },
   }: any = await storefrontPermissions
-    .checkUserPermission()
+    // I don't know why this name sender is necessary, maybe it is some policy or permission.
+    .checkUserPermission('vtex.b2b-orders-history@0.x')
```
mairatma, line 42: This actually seems like a copy/paste problem. The sender is supposed to be the app sending the request, and it's used for 2 things:

- Metrics - tracking which apps makes requests to which other apps
- App access - making sure callers can only call apps that they directly depend on in their `manifest.json` file.

So to be correct this should always be `vtex.b2b-organizations-graphql` here, as you had before. If this causes any permission problem it's because the manifest.json for this repo is missing store-permissions, which shouldn't be the case since there are other callers passing the sender correctly.

node/clients/storefrontPermissions.ts
```diff
@@ -309,6 +304,7 @@ export default class StorefrontPermissions extends AppGraphQLClient {
     return {
       VtexIdclientAutCookie:
         ctx.storeUserAuthToken ?? ctx.adminUserAuthToken ?? ctx.authToken,
+      cookie: `VtexIdclientAutCookie=${this.context.authToken}`,
```
mairatma, line 307: Shouldn't this use the same value of the line above?

graphql/schema.graphql
Rudge: The problem with VBase was another that we fixed. Even if it calls checkPermissions too many times, the problem never occurs again.

About the cache, is it the same for any account? How is the time considered for expiration?

graphql/schema.graphql
Rudge: Yes. Are the directives executed in sequence?

mairatma: Yep!

node/resolvers/Routes/index.ts
mairatma: Got it! Seems like there are other checks then, probably due to the token also be associated with an app. Let's keep orders-history here in this case, though it's weird.

node/resolvers/Routes/index.ts
```diff
@@ -38,6 +38,7 @@ const getUserAndPermissions = async (ctx: Context) => {
   const {
     data: { checkUserPermission },
   }: any = await storefrontPermissions
+    // I don't know why this name sender is necessary, maybe it is some policy or permission.
```
enzomerca, line 41: nit: Can we update this comment with the information on the PR discussion? might be useful in the future.

enzomerca (dismissed):
Just one small comment but LGTM

enzomerca (approved):

PR #141, opened 2023-12-08
feat: add token validation for storefront permissions dependencies

Rudge:

We found some apps that don't pass the token to call graphql queries or mutations. So we need a guarantee to have all integrations send the token, and we will block unauthenticated requests from the app provider.

For the b2b-organization we needed to split the changes into two PRs, one to send token in each storefront-permission integration and validate token for operations that storefront-permissions do not use, and another to put the token validation for operations used by storefront-permissions.

#### How to test it?

[Workspace](https://security--b2bsuite.myvtex.com/)

#### Related to / Depends on

This PR depends on [PR-140](https://github.com/vtex-apps/b2b-organizations-graphql/pull/140)

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -25,8 +25,8 @@ type Query {
     @cacheControl(scope: PRIVATE)
 
   getOrganizationById(id: ID): Organization
-    @cacheControl(scope: PUBLIC, maxAge: SHORT)
-    @auditAccess
+    @cacheControl(scope: PRIVATE, maxAge: SHORT)
+    @checkUserAccess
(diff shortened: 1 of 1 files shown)


enzomerca (approved):

Rudge:
We can’t work with token validation for these operations used by storefront-permissions. It is used in the get session and the storeUserAuthToken is unavailable. So we close this task and don’t merge the changes.

Rudge:
We can’t work with token validation for these operations used by storefront-permissions. It is used in the get session and the storeUserAuthToken is unavailable. So we close this task and don’t merge the changes.

PR #145, opened 2024-02-07
fix: add retry to sendMetric to avoid connection errors

enzomerca:

We are currently seeing a lot of `Error to send metrics from auth metric` on opensearch logs with the following details:
```
...
"method":"post",
"timeout":0,
"url":"https://rc.vtex.com/api/analytics/schemaless-events"},
"request":{},
"isAxiosError":true,
"name":"Error",
"message":"Client network socket disconnected before secure TLS connection was established",
...
```

This PR adds retries to this post call to mitigate connection errors.

#### How to test it?
Change an organization detail on admin and check a metric for updating organization is sent
[Workspace](https://b2bteam1558--b2bstoreqa.myvtex.com/admin)

### DIFF
--- a/node/utils/metrics/metrics.ts
+++ b/node/utils/metrics/metrics.ts
@@ -1,6 +1,8 @@
 import axios from 'axios'
 
 const ANALYTICS_URL = 'https://rc.vtex.com/api/analytics/schemaless-events'
+const MAX_RETRIES = 2
+const RETRY_INTERVAL = 1000 // 1 second
 
 export const B2B_METRIC_NAME = 'b2b-suite-buyerorg-data'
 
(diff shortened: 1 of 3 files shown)


node/utils/metrics/metrics.ts
```diff
@@ -1,6 +1,8 @@
 import axios from 'axios'
 
 const ANALYTICS_URL = 'https://rc.vtex.com/api/analytics/schemaless-events'
```
Rudge, line 3: One thing that I saw in the internal-docs, it is the host for ingestion data. We are using rc.vtex but we should use analytics.vtex as [doc](https://internal-docs.vtex.com/Analytics/VTEX-Data-Platform/data-ingestion-API/#schemaless-events-api). After that, I asked for analytics-support and they confirmed the host in the [thread](https://vtex.slack.com/archives/C02T3NYGVNX/p1699469093595349).

Rudge (approved):

node/utils/metrics/metrics.ts
enzomerca: As discussed offline, apparently "normal" IO apps do not have access to `analytics.vtex.com`, only IO Infra apps.
We decided to keep this URL for now and if we keep seeing errors, we can reachout to #analytics-support and see what can be done.

PR #151, opened 2024-03-19
feat: Add new removeUserWithEmail graphql API for bulk import use case

pabloppupulin:

- Add new removeUserWithEmail graphql API for bulk import use case

### DIFF
--- a/node/resolvers/Mutations/Users.ts
+++ b/node/resolvers/Mutations/Users.ts
@@ -1,6 +1,5 @@
 import { MessageSFPUserAddError, StatusAddUserError } from '../../constants'
 import GraphQLError, { getErrorMessage } from '../../utils/GraphQLError'
-
 import type { ImpersonateMetricParams } from '../../utils/metrics/impersonate'
 import {
   sendImpersonateB2BUserMetric,
@@ -369,7 +368,6 @@ const Users = {
     )
   },
 
-
   removeUserWithEmail: async (
     _: void,
     { orgId, costId, email }: UserArgs,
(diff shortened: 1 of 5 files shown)


node/resolvers/Mutations/Users.ts
```diff
@@ -368,6 +368,66 @@ const Users = {
...
+    } = ctx as any
+
+    const variables = {
+      organizationId: orgId,
+      costCenterId: costId,
+      search: email,
+    }
+
+    storefrontPermissionsClient
+      .listUsersPaginated(variables)
```
enzomerca, line 388: I wonder if there is a way to get a specific user by email/cc id/org id, without the need to list all users. I don't remember if the storefrontPermissions has this API, but if not, maybe we should consider implementing it there (if technically possible) to avoid running into performance issues and edge cases.

node/resolvers/Mutations/Users.ts
```diff
@@ -368,6 +368,66 @@ const Users = {
...
+      .then((result: any) => {
+        const user = result.data.listUsersPaginated.data.find(
+          (u: any) => u.email === email
```
enzomerca, line 391: what happens if the `listUsersPaginated` returns a valid response but there is no match? The API will return 200? Should we return an error maybe, indicating the user does not exists for that org/cc?

pabloppupulin: I left it returning 200, because if the user does not exist for that cost center or organization, then technically it is ok, with the request related.

enzomerca (approved):

enzomerca (approved):

PR #156, opened 2024-05-03
chore: add metric to check access directives

enzomerca:

Add metric to check access directives. With these metrics we can collect data about how these APIs are being used/called.

### Testing
Data example collected from redshift:

`{"name":"b2b-suite-buyerorg-data","account":"b2bstoreqa","fields":{"caller":"vtex.graphql-server@1.67.8","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36","forwardedHost":"testsec--b2bstoreqa.myvtex.com","hasAdminToken":true,"hasApiToken":false,"hasStoreToken":false,"operation":"getCostCenterByIdStorefront"},"kind":"b2b-organizations-graphql-auth-event","description":"CheckUserAccess"}`

### DIFF
--- a/node/resolvers/directives/checkAdminAccess.ts
+++ b/node/resolvers/directives/checkAdminAccess.ts
@@ -21,15 +21,17 @@ export class CheckAdminAccess extends SchemaDirectiveVisitor {
       } = context
 
       // send metric to collect data about caller and tokens used in the request
-      const operation = field.astNode?.name?.value ?? context.request.url
+      const operation = field.astNode?.name?.value ?? context?.request?.url
       const metric = new AuthMetric(
-        context.vtex.account,
+        context?.vtex?.account,
         {
-          caller: context.request.header['x-vtex-caller'] as string,
-          userAgent: context.request.header['user-agent'] as string,
-          forwardedHost: context.request.header['x-forwarded-host'] as string,
+          caller: context?.request?.headers['x-vtex-caller'] as string,
+          userAgent: context?.request?.headers['user-agent'] as string,
+          forwardedHost: context?.request?.headers[
+            'x-forwarded-host'
+          ] as string,
           hasAdminToken: !!adminUserAuthToken,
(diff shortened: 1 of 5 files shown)


Rudge (dismissed):

node/resolvers/directives/checkAdminAccess.ts
```diff
@@ -18,6 +20,24 @@ export class CheckAdminAccess extends SchemaDirectiveVisitor {
         clients: { identity },
       } = context
 
+      // send metric to collect data about caller and tokens used in the request
+      const operation = field.astNode?.name?.value ?? context.request.url
```
fredppm, line 24: 🟢 
Scripts are not my favorite languages.
Suggest to add` ?` In all fields navigation to avoid a a code break because of metrics.

node/resolvers/directives/checkAdminAccess.ts
```diff
@@ -18,6 +20,24 @@ export class CheckAdminAccess extends SchemaDirectiveVisitor {
...
+      const metric = new AuthMetric(
+        context.vtex.account,
```
fredppm, line 26: 🤔 
What happens when a header doesn't exist?

fredppm (dismissed):
I am not expect of Typescript but I added some suggestions.
Dispite that, the code is good to release ..

node/resolvers/directives/checkAdminAccess.ts
enzomerca: If the `context.request.header` structure exists but a specific header (e.g. `'x-vtex-caller'`) is not there, we would get an `undefined` value, which is fine and expected.
On the other hand, if the `context.request.header` structure does not exist at all, we would have a problem (error: `Cannot read properties of undefined (reading 'x-vtex-caller')`), in which case this code would be broken even before this change :)

Rudge (approved):

PR #159, opened 2024-05-23
chore: improve logging/metric for user/admin access directives

enzomerca:

Improve logging and metrics for checkUserAccess and checkAdminAccess directives.


node/resolvers/directives/helper.ts
```diff
@@ -0,0 +1,202 @@
...
+  pageSize: 50,
+}
+
+const USER_DATA_ENTITY = 'b2b_users'
+const USER_DATA_ENTITY_SCHEMA = 'v0.1.2'
+
+// This function checks if given email is an user part of a buyer org.
+// This was copied (slightly modified) from storefront-permissions to avoid
+// a call to other application just to check if the user is part of a buyer org.
+const isUserPartOfBuyerOrg = async (email: string, ctx: Context) => {
```
Rudge, line 13: It is better to move this function to another file, something like b2b-users.

Rudge (dismissed):

Rudge (approved):

PR #161, opened 2024-05-29
fix: check access directive

enzomerca:

Fix check access directives by allowing appkey tokens for authentication


fredppm (approved):
🟢 Good

PR #165, opened 2024-07-10
fix: update admin token validation and add audit metric

enzomerca:

* Add audit metrics to some graphql APIs
* Add license manager client (copied from `storefront-permissions`)
* Improve logging and access control

More info: https://vtex-dev.atlassian.net/browse/B2BTEAM-1746

### DIFF
--- a/node/clients/LMClient.ts
+++ b/node/clients/LMClient.ts
@@ -1,0 +1,30 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { InstanceOptions, IOContext } from '@vtex/api'
+import { ExternalClient } from '@vtex/api'
+
+export default class LMClient extends ExternalClient {
+  constructor(ctx: IOContext, options?: InstanceOptions) {
+    super(`http://${ctx.account}.vtexcommercestable.com.br/`, ctx, {
+      ...options,
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        VtexIdclientAutCookie: ctx.authToken,
+      },
+    })
+  }
+
+  public getUserAdminPermissions = async (account: string, userId: string) => {
+    return this.get(
+      `/api/license-manager/pvt/accounts/${encodeURI(
+        account
+      )}/logins/${encodeURI(userId)}/granted`
+    ).then((res: any) => {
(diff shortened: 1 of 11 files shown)


node/resolvers/directives/checkAdminAccess.ts
```diff
@@ -48,13 +58,15 @@ export class CheckAdminAccess extends SchemaDirectiveVisitor {
           hasApiToken,
           hasValidApiToken,
           hasStoreToken,
```
Matheus-Aguilar, line 60: Since the store token is not used in this directive, we should remove it for better code readability.

Matheus-Aguilar (approved):
LGTM. Only a few minor comments.

node/resolvers/directives/checkAdminAccess.ts
enzomerca: we have it here just to have a metric to identify if this token is being provided incorrectly for this case.

node/clients/LMClient.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { InstanceOptions, IOContext } from '@vtex/api'
+import { ExternalClient } from '@vtex/api'
+
+export default class LMClient extends ExternalClient {
+  constructor(ctx: IOContext, options?: InstanceOptions) {
+    super(`http://${ctx.account}.vtexcommercestable.com.br/`, ctx, {
```
fredppm, line 7: 🟡 This endpoint is public, and the HTTP is not secure ( HTTPS ), so the token is easily intercepted on the internet.
If possible use an internal endpoint or use HTTPS

node/resolvers/directives/helper.ts
```diff
@@ -126,8 +143,40 @@ export const validateStoreToken = async (
...
+        message: 'Error validating store token:',
+        err,
+      })
     }
   }
 
   return { hasStoreToken, hasValidStoreToken, hasCurrentValidStoreToken }
 }
+
+export const validateAdminTokenOnHeader = async (
```
fredppm, line 156: 🟡 All other validation methods are wrapped by a try-catch. Why is this not?

PR #168, opened 2024-07-30
feat: add paging support to getSellers query

icaroov:

The `getSellersPaginated` query now accepts a `paging` input parameter, allowing the client to specify the range of sellers to retrieve. The response includes a `paging` object with information about the total number of sellers and the range of sellers returned. This change improves the flexibility and efficiency of retrieving seller data.

#### How to test it?

```graphql
query GetSellersPaginated($page: Int, $pageSize: Int) {
  getSellersPaginated(page: $page, pageSize: $pageSize) {
    pagination {
      page
      pageSize
      total
    }
    items {
      id
      name
      email
    }
  }
}

```

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -110,25 +110,25 @@ type Query {
     @cacheControl(scope: PUBLIC, maxAge: SHORT)
     @auditAccess
   getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
-  getSellersPaginated(paging: PaginationInput): GetSellersResponse
-    @checkAdminAccess
-    @cacheControl(scope: PRIVATE)
+  getSellersPaginated(
+    args: GetSellersPaginatedArgsInput
+  ): GetSellersPaginatedResponse @checkAdminAccess @cacheControl(scope: PRIVATE)
 }
 
-input PaginationInput {
-  from: Int
-  to: Int
+input GetSellersPaginatedArgsInput {
+  page: Int
+  pageSize: Int
 }
 
 type PaginationResponse {
-  from: Int
-  to: Int
+  page: Int
+  pageSize: Int
   total: Int
 }
 
-type GetSellersResponse {
+type GetSellersPaginatedResponse {
   items: [Seller]
-  paging: PaginationResponse
+  pagination: PaginationResponse
 }
 
 type Mutation {
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -111,7 +111,7 @@ type Query {
     @auditAccess
   getSellers: [Seller] @validateAdminUserAccess @cacheControl(scope: PRIVATE)
   getSellersPaginated(page: Int, pageSize: Int): GetSellersPaginatedResponse
-    @checkAdminAccess
+    @validateAdminUserAccess
     @cacheControl(scope: PRIVATE)
 }
 
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -110,14 +110,9 @@ type Query {
     @cacheControl(scope: PUBLIC, maxAge: SHORT)
     @auditAccess
   getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
-  getSellersPaginated(
-    args: GetSellersPaginatedArgsInput
-  ): GetSellersPaginatedResponse @checkAdminAccess @cacheControl(scope: PRIVATE)
-}
-
-input GetSellersPaginatedArgsInput {
-  page: Int
-  pageSize: Int
+  getSellersPaginated(page: Int, pageSize: Int): GetSellersPaginatedResponse
+    @checkAdminAccess
+    @cacheControl(scope: PRIVATE)
 }
 
 type PaginationResponse {
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -109,7 +109,8 @@ type Query {
   getB2BSettings: B2BSettings
     @cacheControl(scope: PUBLIC, maxAge: SHORT)
     @auditAccess
-  getSellers(paging: PaginationInput): GetSellersResponse
+  getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
+  getSellersPaginated(paging: PaginationInput): GetSellersResponse
     @checkAdminAccess
     @cacheControl(scope: PRIVATE)
 }
@@ -127,7 +128,7 @@ type PaginationResponse {
 
 type GetSellersResponse {
   items: [Seller]
-  pagination: PaginationResponse
+  paging: PaginationResponse
 }
 
 type Mutation {
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -109,7 +109,25 @@ type Query {
   getB2BSettings: B2BSettings
     @cacheControl(scope: PUBLIC, maxAge: SHORT)
     @auditAccess
-  getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
+  getSellers(paging: PaginationInput): GetSellersResponse
+    @checkAdminAccess
+    @cacheControl(scope: PRIVATE)
+}
+
+input PaginationInput {
+  from: Int
+  to: Int
+}
+
+type PaginationResponse {
+  from: Int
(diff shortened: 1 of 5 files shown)


enzomerca (changes requested):

graphql/schema.graphql
```diff
@@ -109,7 +109,25 @@ type Query {
   getB2BSettings: B2BSettings
     @cacheControl(scope: PUBLIC, maxAge: SHORT)
     @auditAccess
-  getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
+  getSellers(paging: PaginationInput): GetSellersResponse
+    @checkAdminAccess
+    @cacheControl(scope: PRIVATE)
```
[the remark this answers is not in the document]
icaroov: It makes sense, I can do this!

node/resolvers/Queries/Settings.ts
```diff
@@ -47,12 +47,33 @@ const B2BSettings = {
...
+  ) => {
     const {
       clients: { sellers },
     } = ctx
 
-    return (await sellers.getSellers())?.items
+    const { paging } = args
+
+    try {
+      return sellers.getSellers(paging)
```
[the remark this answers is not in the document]
icaroov: Done!

graphql/schema.graphql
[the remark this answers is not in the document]
icaroov: Done!

node/clients/sellers.ts
```diff
@@ -19,7 +19,16 @@ export default class SellersClient extends JanusClient {
     })
   }
 
-  public async getSellers(): Promise<{ items: Seller[] }> {
-    return this.http.get(SELLERS_PATH)
+  public async getSellers(paging?: {
+    from: number
+    to: number
+  }): Promise<{ items: Seller[] }> {
```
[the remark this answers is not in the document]
icaroov: Done!

node/clients/sellers.ts
```diff
@@ -19,7 +39,69 @@ export default class SellersClient extends JanusClient {
...
+          return isNumber ? 1 : 0
+        },
+        (seller: Seller) => {
+          const name = seller.name ?? ''
+
+          return isNaN(name) ? name : Number(name)
+        },
+      ],
+      ['asc', 'asc']
+    )
```
enzomerca, line 105: can you add a comment to the code explaining a bit how this sort works?

node/resolvers/Queries/Settings.ts
```diff
@@ -52,7 +53,38 @@ const B2BSettings = {
...
+    } catch (e) {
+      if (e.message) {
+        throw new GraphQLError(e.message)
+      } else if (e.response?.data?.message) {
+        throw new GraphQLError(e.response.data.message)
+      } else {
+        throw new GraphQLError(e)
+      }
+    }
+  },
```
enzomerca, line 67: as we are creating a new method, I think it is better if we do not touch this part of the code (to avoid changing behaviour of the existing method)

I know this is improving the code, but still, it could cause unexpected behaviour for current use cases.

enzomerca (changes requested):

graphql/schema.graphql
```diff
@@ -110,6 +110,25 @@ type Query {
     @cacheControl(scope: PUBLIC, maxAge: SHORT)
     @auditAccess
   getSellers: [Seller] @checkAdminAccess @cacheControl(scope: PRIVATE)
+  getSellersPaginated(
+    args: GetSellersPaginatedArgsInput
```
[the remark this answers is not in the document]
icaroov: Sure!

node/clients/sellers.ts
```diff
@@ -19,7 +39,69 @@ export default class SellersClient extends JanusClient {
...
+        from: number
+        to: number
+        total: number
+      }
+      items: Seller[]
+    }>(SELLERS_PATH, {
+      metric: 'sellers-get',
+      params: {
+        from: argsWithDefaults?.page ?? INITIAL_LIST_OPTIONS.page,
+        to: argsWithDefaults?.pageSize ?? INITIAL_LIST_OPTIONS.pageSize,
```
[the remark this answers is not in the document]
icaroov: That's correct, but the logic you used was being done on the client, but now I've reversed it and it only works on the backend now.

node/clients/sellers.ts
icaroov: Done!

node/clients/sellers.ts
```diff
@@ -19,7 +39,69 @@ export default class SellersClient extends JanusClient {
...
+      },
+    })
+
+    if (!result.items) {
+      return {
+        items: [],
+        pagination: {
+          page: 0,
```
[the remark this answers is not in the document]
icaroov: Adjusted!

node/clients/sellers.ts
```diff
@@ -19,7 +39,76 @@ export default class SellersClient extends JanusClient {
     })
   }
 
-  public async getSellers(): Promise<{ items: Seller[] }> {
-    return this.http.get(SELLERS_PATH)
+  public async getSellers(opts?: GetSellersOpts): Promise<GetSellersResponse> {
```
enzomerca, line 42: Can we keep the original `getSellers` implementation and create a new `getSellersPaginated`? The `getSellers` original implementation is very simple and small, is not like we are duplicating a lot of code. This way we keep the original behavior without any changes.

node/clients/sellers.ts
```diff
@@ -19,7 +39,76 @@ export default class SellersClient extends JanusClient {
...
+
+    const result = await this.http.get<{
+      paging: {
+        from: number
+        to: number
+        total: number
+      }
+      items: Seller[]
+    }>(SELLERS_PATH, {
+      metric: 'sellers-get',
```
enzomerca, line 56: can we change `sellers-get` to `sellers-paginated-get`?

node/clients/sellers.ts
```diff
@@ -19,7 +39,76 @@ export default class SellersClient extends JanusClient {
...
+      params: {
+        from: (argsWithDefaults.page - 1) * argsWithDefaults.pageSize,
```
enzomerca, line 58: what if `page = 0`? what happens? do we need to protect against this case?
I also wonder what happens if we provide other invalid values like `from < to`.
Can you check how the sellers API behave when such cases happens? Maybe we can let the sellers API handle it, or we might need to check these values before sending it to the sellers API.

icaroov: Good point! The API returns a 400 bad request error in both cases, without any message explaining what caused the error. I'm going to return an exception in the case of `from` and `to` to make it clear to whoever is using it what caused the error.

node/clients/sellers.ts
```diff
@@ -20,6 +35,98 @@ export default class SellersClient extends JanusClient {
   }
 
   public async getSellers(): Promise<{ items: Seller[] }> {
-    return this.http.get(SELLERS_PATH)
+    return this.http.get(SELLERS_PATH, {
+      metric: 'sellers-get',
```
enzomerca, line 39: nit: lets not touch this code

node/clients/sellers.ts
```diff
@@ -20,6 +35,98 @@ export default class SellersClient extends JanusClient {
...
+    const page = Math.max(1, argsWithDefaults.page)
+
+    const from = (page - 1) * pageSize
+    const to = page * pageSize
+
+    if (from >= to) {
+      throw new Error(
+        'Invalid pagination values: `from` should be less than `to`. Please make sure you are passing valid values for `page` and `pageSize`.'
+      )
+    }
```
enzomerca, line 67: now that I think about it, we only need to make sure `page` and `pageSize` are both `> 0`. If that is the case, `from` will always be smaller than `to`. Makes sense?
If that is the case, we can simplify this to:
```suggestion
    const { page, pageSize } = argsWithDefaults

    if (page < 1 || pageSize < 1) {
      throw new Error(
        'Please make sure you are passing valid values for `page` and `pageSize`.'
      )
    }

    const from = (page - 1) * pageSize
    const to = page * pageSize
```

what do you think?

icaroov: Great, makes sense!

enzomerca (approved):
LGTM!
I will do some tests later but we should merge/deploy this change next week!

PR #172, opened 2024-08-29
Add new query getOrganizationsByEmailPaginated

vhaalmeida:

This new query will allow to retrieve the users' organizations paginated.
Currently when an user is associated to a big number of organizations is generating a timeout.

This PR should be merged only when this other one is merged: https://github.com/vtex-apps/storefront-permissions/pull/157

#### How to test it?

Access https://timeoutsfp--jamalstore.myvtex.com/admin/graphql-ide
Run the query below in the b2b-organizations-graphql app

```
{
  getOrganizationsByEmailPaginated(
    email: "victor.almeida+2@vtex.com",
    page: 1
    pageSize: 25
  ) {
    data {
      id
      clId
     	costId
      orgId
      organizationName
      costCenterName
    }
    pagination {
      page
      pageSize
      total
    }
  }
}
```

[Workspace](https://timeoutsfp--jamalstore.myvtex.com/admin/graphql-ide)

image

#### Related to / Depends on

https://github.com/vtex-apps/storefront-permissions/pull/157

### DIFF
--- a/node/clients/storefrontPermissions.ts
+++ b/node/clients/storefrontPermissions.ts
@@ -8,6 +8,7 @@ import saveUser from '../mutations/saveUser'
 import updateUser from '../mutations/updateUser'
 import getB2BUser from '../queries/getB2BUser'
 import getOrganizationsByEmail from '../queries/getOrganizationsByEmail'
+import getOrganizationsByEmailPaginated from '../queries/getOrganizationsByEmailPaginated'
 import getPermission from '../queries/getPermission'
 import getRole from '../queries/getRole'
 import getUser from '../queries/getUser'
@@ -36,6 +37,18 @@ export default class StorefrontPermissions extends AppGraphQLClient {
       query: getOrganizationsByEmail,
       variables: {
         email,
+      },
+    })
+  }
+
+  public getOrganizationsByEmailPaginated = async (email: string, page: number, pageSize: number): Promise<any> => {
+    return this.query({
+      extensions: this.getPersistedQuery(),
+      query: getOrganizationsByEmailPaginated,
+      variables: {
+        email,
+        page,
+        pageSize,
       },
     })
   }
--- a/node/resolvers/Queries/Organizations.ts
+++ b/node/resolvers/Queries/Organizations.ts
@@ -230,8 +230,109 @@ const Organizations = {
       )
     })
 
+    console.log(`organizations`, organizations)
+
     try {
       return organizations
+    } catch (error) {
+      logger.error({
+        error,
+        message: 'getOrganizationsByEmail-error',
+      })
+      throw new GraphQLError(getErrorMessage(error))
+    }
+  },
+
+  getOrganizationsByEmailPaginated: async (
+    _: void,
+    {
+      email,
+      page = 1,
+      pageSize = 25,
+    }: {
+      email: string
+      page: number
+      pageSize: number
+    },
+    {
+      clients: { storefrontPermissions, session },
+      vtex: { logger, sessionToken, adminUserAuthToken },
+    }: any
+  ) => {
+
+    const organizationFilters: string[] = []
+    let fromSession = false
+    const {
+      data: { checkUserPermission },
+    }: any = await storefrontPermissions
+      .checkUserPermission('vtex.b2b-organizations@1.x')
+      .catch((error: any) => {
+        logger.error({
+          error,
+          message: 'checkUserPermission-error',
+        })
+
+        return {
+          data: {
+            checkUserPermission: null,
+          },
+        }
+      })
+
+    if (
+      (!adminUserAuthToken &&
+        !checkUserPermission?.permissions.includes('add-sales-users-all')) ||
+      !(email?.length > 0)
+    ) {
+      const sessionData = await session
+        .getSession(sessionToken as string, ['*'])
+        .then((currentSession: any) => {
+          return currentSession.sessionData
+        })
+        .catch((error: any) => {
+          logger.warn({
+            error,
+            message: 'getOrganizationsByEmail-session-error',
+          })
+
+          return null
+        })
+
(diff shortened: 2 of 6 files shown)


wender (changes requested):
Please run the Lint in the files changed to auto-fix the line-break issues?
If `vtex setup` don't install all dependencies, you need to run `yarn` in the root and node folder

wender (approved):

wender (approved):
LGTM

node/clients/storefrontPermissions.ts
```diff
@@ -40,6 +41,22 @@ export default class StorefrontPermissions extends AppGraphQLClient {
     })
   }
 
+  public getOrganizationsByEmailPaginated = async (
```
enzomerca, line 44: Do we need this function here and on the storefront-permissions app? ([PR](https://github.com/vtex-apps/storefront-permissions/pull/157#pullrequestreview-2280369421))

Can't we query Master Data directly from this app and avoid additional call to another app?

node/resolvers/Queries/Organizations.ts
```diff
@@ -241,6 +241,111 @@ const Organizations = {
...
+        return {
+          data: {
+            checkUserPermission: null,
+          },
+        }
+      })
+
+    if (
+      (!adminUserAuthToken &&
+        !checkUserPermission?.permissions.includes('add-sales-users-all')) ||
```
enzomerca, line 281: why are we checking if the user has `add-sales-users-all` permission if this is just a `getOrganizationsByEmail` query?

node/resolvers/Queries/Organizations.ts
```diff
@@ -241,6 +241,111 @@ const Organizations = {
...
+        .catch((error: any) => {
+          logger.warn({
+            error,
+            message: 'getOrganizationsByEmail-session-error',
+          })
+
+          return null
+        })
+
+      if (checkUserPermission?.permissions.includes('add-users-organization')) {
```
enzomerca, line 298: why are we checking if the user has `add-users-organization` permission if this is just a `getOrganizationsByEmail` query?

node/resolvers/Queries/Organizations.ts
```diff
@@ -241,6 +241,111 @@ const Organizations = {
...
+        const orgId =
+          sessionData?.namespaces?.['storefront-permissions']?.organization
+            ?.value
+
+        if (!orgId) {
+          throw new Error('No permission for getting the organizations')
+        }
+
+        organizationFilters.push(orgId)
```
enzomerca, line 307: session will have just one org id, the current logged org, right? and below we are filtering the result to return only the orgs added to `organizationsFilters`, so the `getOrganizationsByEmail` would not return all the organizations for the given email in this case, right?

node/resolvers/Queries/Organizations.ts
```diff
@@ -241,6 +241,111 @@ const Organizations = {
...
+      }
+
+      if (!(email?.length > 0)) {
+        email = sessionData?.namespaces?.profile?.email?.value
+        fromSession = true
+      }
+    }
```
enzomerca, line 314: the function is named `getOrganizationsByEmail` but we are allowing the user to not provide an email and get the email from the session. Who ever is going to call this API can do that on their side, right?

I prefer to remove this, make the email a required field and simplify the code.

enzomerca (changes requested):
A general comment: I would try to avoid using session data as much as possible in these new APIs as this would make it harder to use it in a headless scenario and in integrations where the system has no session token to provide on the call. (we can discuss offline if you have any concerns by removing the session token usage/verification)

node/clients/storefrontPermissions.ts
vhaalmeida: @enzomerca I'm following the same approach of the getOrganizationsByEmail query

node/resolvers/Queries/Organizations.ts
vhaalmeida: Hi @enzomerca 
Is the same approach of the getOrganizationsByEmail query. If I'm not mistaken this is for the case of a query run through the webstore where there's no admin auth cookie.

node/resolvers/Queries/Organizations.ts
vhaalmeida: Hi @enzomerca 
Is the same approach of the getOrganizationsByEmail query. If I'm not mistaken this is for the case of a query run through the webstore where there's no admin auth cookie.

node/resolvers/Queries/Organizations.ts
vhaalmeida: Hi @enzomerca 
Is the same approach of the getOrganizationsByEmail query. This is the case for the request from the storefront where there's no admin auth cookie.

node/resolvers/Queries/Organizations.ts
vhaalmeida: Hi @enzomerca 
Is the same approach of the getOrganizationsByEmail query.

node/clients/storefrontPermissions.ts
Rudge: Yes, but we are removing the cyclic dependencies, and this is one of them, and thinking about the domain and context, the storefront shouldn't know about the organization.

node/resolvers/Queries/Organizations.ts
Rudge: Why don't you remove it to test? It appears to be a bug or old mistake.

node/resolvers/Queries/Organizations.ts
Rudge: @vhaalmeida We would like to avoid old mistakes, and we have a lot of tasks to improve the source code and the B2B suite app implementations. So our advice is to adjust the rules to test.

Rudge (changes requested):

PR #181, opened 2024-10-11
draft: include new validation on validateAdminUserAccess

giurigaud:
#### How to test it?

[Workspace](Link goes here!)

### DIFF
--- a/graphql/directives.graphql
+++ b/graphql/directives.graphql
@@ -1,6 +1,8 @@
 directive @withSession on FIELD | FIELD_DEFINITION
 directive @withPermissions on FIELD | FIELD_DEFINITION
-directive @validateAdminUserAccess on FIELD | FIELD_DEFINITION
+directive @validateAdminUserAccess(
+  orgPermission: String
+) on FIELD | FIELD_DEFINITION
 directive @checkUserAccess on FIELD | FIELD_DEFINITION
 directive @checkAdminAccess on FIELD | FIELD_DEFINITION
 directive @auditAccess on FIELD | FIELD_DEFINITION
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -153,7 +153,8 @@ type Mutation {
     notifyUsers: Boolean
   ): OrganizationCostCenterResponse
     @checkAdminAccess
-    @cacheControl(scope: PRIVATE)
+    @validateAdminUserAccess(orgPermission: "buyer_organization_edit")
+    @cacheControl(scope: PRIVATE, maxAge: 300)
   createOrganizationAndCostCentersWithId(
     input: NormalizedOrganizationInput!
   ): MasterDataResponse @checkAdminAccess @cacheControl(scope: PRIVATE)
--- a/node/clients/LMClient.ts
+++ b/node/clients/LMClient.ts
@@ -1,6 +1,7 @@
+/* eslint-disable no-console */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import type { InstanceOptions, IOContext } from '@vtex/api'
-import { ExternalClient } from '@vtex/api'
+import { ExternalClient, ForbiddenError } from '@vtex/api'
 
 export default class LMClient extends ExternalClient {
   constructor(ctx: IOContext, options?: InstanceOptions) {
@@ -22,6 +23,24 @@ export default class LMClient extends ExternalClient {
     ).then((res: any) => {
       return res
     })
+  }
+
+  public checkUserAdminPermission = async (
+    account: string,
+    userEmail: string,
+    resourceCode: string
+  ) => {
+    const productCode = '97'
+
+    const checkOrgPermission = await this.get<boolean>(
+      `/api/license-manager/pvt/accounts/${account}/products/${productCode}/logins/${userEmail}/resources/${resourceCode}/granted`
+    )
+
+    if (!checkOrgPermission) {
+      throw new ForbiddenError('Unauthorized Access')
+    }
+
+    return checkOrgPermission
   }
 
   public getAccount = async () => {
--- a/node/resolvers/directives/helper.ts
+++ b/node/resolvers/directives/helper.ts
@@ -1,8 +1,11 @@
+/* eslint-disable no-console */
+
 import { isUserPartOfBuyerOrg } from '../Queries/Users'
 
 export const validateAdminToken = async (
   context: Context,
-  adminUserAuthToken: string
+  adminUserAuthToken: string,
+  orgPermission?: 'buyer_organization_edit' | 'buyer_organization_view'
 ): Promise<{
   hasAdminToken: boolean
   hasValidAdminToken: boolean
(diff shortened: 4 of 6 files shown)


node/resolvers/directives/helper.ts
```diff
@@ -44,6 +55,11 @@ export const validateAdminToken = async (
     }
   }
 
+  console.log({ hasValidAdminToken, orgPermission, tokenType })
+  if (hasValidAdminToken && orgPermission && tokenType === 'user') {
+    await lm.checkUserAdminPermission(account, userEmail, orgPermission)
+  }
```
enzomerca, line 61: I think we can move this check to inside the try above, right after the `lm.getUserAdminPermissions` call

node/resolvers/directives/helper.ts
```diff
@@ -79,6 +95,9 @@ export const validateApiToken = async (
         token,
       })
 
+      console.log('validateApiToken')
+      console.log({ authUser })
+      console.log('end validateApiToken')
       // we set this flag to true if the token is valid by current standards
       // in the future we should remove this line
       hasCurrentValidApiToken = true
```
enzomerca, line 103: we need to do the same check below, after the `lm.getUserAdminPermissions`

node/resolvers/directives/helper.ts
```diff
@@ -1,8 +1,11 @@
+/* eslint-disable no-console */
+
 import { isUserPartOfBuyerOrg } from '../Queries/Users'
 
 export const validateAdminToken = async (
   context: Context,
-  adminUserAuthToken: string
+  adminUserAuthToken: string,
+  orgPermission?: 'buyer_organization_edit' | 'buyer_organization_view'
```
enzomerca, line 8: sorry, my typescript is a bit rusty. What does this means exactly?
The `orgPermission` is optional, and should be either `buyer_organization_edit` or `buyer_organization_view`? Otherwise it will be empty? Is that it?

node/clients/LMClient.ts
```diff
@@ -24,6 +25,24 @@ export default class LMClient extends ExternalClient {
     })
   }
 
+  public checkUserAdminPermission = async (
+    account: string,
+    userEmail: string,
+    resourceCode: string
+  ) => {
+    const productCode = '97'
```
enzomerca, line 33: lets make this a constant in the file and add a comment with the resource name, something like this:
`const B2B_LM_PRODUCT_CODE = '97' // resource name on lincense manager = B2B`

enzomerca (commented):
Nice! I think that is the idea! I just left some small comments.

node/resolvers/directives/helper.ts
giurigaud: Yes, if `org Permission` came, it should be `buyer_organization_edit` or `buyer organization view`, otherwise the property will be `undefined`

node/clients/LMClient.ts
giurigaud: niceee, much better

node/resolvers/directives/helper.ts
giurigaud: the problem with passing this to the try catch above is that the catch does not block the call, it just logs the error

graphql/schema.graphql
```diff
@@ -24,14 +25,16 @@ type Query {
     sortedBy: String = "name"
   ): OrganizationResult
     @cacheControl(scope: PRIVATE, maxAge: SHORT)
-    @validateStoreUserAccess
+    @validateAdminUserAccess(orgPermission: "buyer_organization_view")
+    @validateStoreUserAccess(orgPermission: "buyer_organization_view")
```
enzomerca, line 29: We should have just one validate directive per API (in this case we should keep only the `@validateStoreUserAccess`

graphql/directives.graphql
```diff
@@ -1,7 +1,11 @@
 directive @withSession on FIELD | FIELD_DEFINITION
 directive @withPermissions on FIELD | FIELD_DEFINITION
-directive @validateAdminUserAccess on FIELD | FIELD_DEFINITION
-directive @validateStoreUserAccess on FIELD | FIELD_DEFINITION
+directive @validateAdminUserAccess(
+  orgPermission: String
```
enzomerca, line 4: I think we need to change `orgPermission` to something more meaningful... maybe `requiredPermission`?

graphql/schema.graphql
```diff
@@ -11,6 +11,7 @@ type Query {
     sortedBy: String = "created"
   ): OrganizationRequestResult
     @cacheControl(scope: PRIVATE, maxAge: SHORT)
+    @validateAdminUserAccess(orgPermission: "buyer_organization_view")
```
enzomerca, line 14: for now lets not add the directive to APIs that don't have it. Lets just add the parameter to APIs that already have either `validateAdminUserAccess` or `validateStoreUserAccess`.

node/clients/LMClient.ts
```diff
@@ -24,6 +26,24 @@ export default class LMClient extends ExternalClient {
     })
   }
 
+  public checkUserAdminPermission = async (
+    account: string,
+    userEmail: string,
+    resourceCode: string
+  ) => {
+    const productCode = B2B_LM_PRODUCT_CODE // resource name on lincense manager = B2B
```
enzomerca, line 34: the actual name of this resource is `Buyer Organization` actually (I gave you the wrong info before). So lets update the comment here
```suggestion
    const productCode = B2B_LM_PRODUCT_CODE // resource name on lincense manager = Buyer Organization
```

node/clients/LMClient.ts
```diff
@@ -24,6 +26,24 @@ export default class LMClient extends ExternalClient {
     })
   }
 
+  public checkUserAdminPermission = async (
```
enzomerca, line 29:
```suggestion
  public checkAdminUserRequiredPermission = async (
```

node/clients/LMClient.ts
```diff
@@ -24,6 +26,24 @@ export default class LMClient extends ExternalClient {
...
+
+    const checkOrgPermission = await this.get<boolean>(
+      `/api/license-manager/pvt/accounts/${account}/products/${productCode}/logins/${userEmail}/resources/${resourceCode}/granted`
+    )
+
+    if (!checkOrgPermission) {
+      throw new ForbiddenError('Unauthorized Access')
+    }
+
+    return checkOrgPermission
```
enzomerca, line 44:
```suggestion
    const checkRequiredPermission = await this.get<boolean>(
      `/api/license-manager/pvt/accounts/${account}/products/${productCode}/logins/${userEmail}/resources/${resourceCode}/granted`
    )

    if (!checkRequiredPermission) {
      throw new ForbiddenError('Unauthorized Access')
    }

    return checkRequiredPermission
```

node/resolvers/directives/helper.ts
```diff
@@ -2,7 +2,8 @@ import { isUserPartOfBuyerOrg } from '../Queries/Users'
...
```
enzomerca, line 6: (lets change `orgPermission` everywhere to `requiredPermission`
```suggestion
  requiredPermission?: 'buyer_organization_edit' | 'buyer_organization_view'
```

node/utils/constants.ts
```diff
@@ -14,3 +14,5 @@ export const ORGANIZATION_REQUEST_STATUSES = {
 export const MARKETING_TAGS = {
   VBASE_BUCKET: 'b2b_marketing_tags',
 }
+
+export const B2B_LM_PRODUCT_CODE = '97' // resource name on lincense manager = B2B
```
enzomerca, line 18: update comment here as well.

graphql/schema.graphql
giurigaud: but if we don't add in getOrganizations for example the permissions will pass. Will it really be like that?

graphql/directives.graphql
giurigaud: I agree

PR #184, opened 2024-10-18
fix: [KI 1092989] getOrganizationsByEmail to return only active organizations

giurigaud:
Inactive organizations shown when changing organizations

#### How to test it?

- Register at least two organizations with the same email;
- Inactivate an organization;
- Enter the store and change the organization; the inactive organization can't be shown in the list.

[Workspace](https://giurigaud--b2bstore005.myvtex.com/)

Before change:
KI 1092989  Before

After change:
Screenshot 2024-10-18 at 12 23 14

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -96,6 +96,7 @@ type Query {
   getActiveOrganizationsByEmail(email: String): [B2BOrganization]
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
+    @validateStoreUserAccess
 
   checkOrganizationIsActive(id: String): Boolean
     @cacheControl(scope: PRIVATE)
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -93,6 +93,10 @@ type Query {
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
 
+  getActiveOrganizationsByEmail(email: String): [B2BOrganization]
+    @checkUserAccess
+    @cacheControl(scope: PRIVATE)
+
   checkOrganizationIsActive(id: String): Boolean
     @cacheControl(scope: PRIVATE)
     @auditAccess
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -94,7 +94,6 @@ type Query {
     @cacheControl(scope: PRIVATE)
 
   getActiveOrganizationsByEmail(email: String): [B2BOrganization]
-    @checkUserAccess
     @cacheControl(scope: PRIVATE)
     @validateStoreUserAccess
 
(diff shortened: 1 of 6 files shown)


node/typings.d.ts
```diff
@@ -163,6 +163,15 @@ interface Organization {
...
 
+export type GetOrganizationsByEmailWithStatus = {
+  costId: string
+  orgId: string
+  roleId: string
+  id: string
+  clId: string
+  status: string
+}
+
```
enzomerca, line 174: I'm not sure this should be in this file... or should it be an `interface` like the other cases here?

node/resolvers/Queries/Organizations.ts
```diff
@@ -182,11 +187,13 @@ const Organizations = {
   getOrganizationsByEmail: async (
```
enzomerca, line 187: I'm wondering if we should create another function called `getActiveOrganizationsByEmail`, so we don't change the behavior of this function... so the caller has 2 options:
1. call `getOrganizationsByEmail` to get all organizations for that email (either active or inactive).
2. call `getActiveOrganizationsByEmail` to get all/only active organizations for that email.

giurigaud: sure, done

node/typings.d.ts
giurigaud: done

graphql/schema.graphql
```diff
@@ -93,6 +93,10 @@ type Query {
     @checkUserAccess
     @cacheControl(scope: PRIVATE)
 
+  getActiveOrganizationsByEmail(email: String): [B2BOrganization]
+    @checkUserAccess
```
enzomerca, line 97: As this is a new API, lets use `validateStoreUserAccess` instead

node/resolvers/Queries/Organizations.ts
```diff
@@ -265,6 +278,49 @@ const Organizations = {
...
+    const activeOrganizations = organizationsWithStatus.filter(
+      (organization) => organization.status === 'active'
+    )
+
+    try {
+      return activeOrganizations
+    } catch (error) {
+      logger.error({
+        error,
+        message: 'getOrganizationsByEmail-error',
```
enzomerca, line 318:
```suggestion
        message: 'getActiveOrganizationsByEmail-error',
```

node/resolvers/Queries/Organizations.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-await-in-loop */
```
enzomerca, line 1: do we need to disable this lint?

node/resolvers/Queries/Organizations.ts
```diff
@@ -9,6 +10,16 @@ import {
 import type { Organization } from '../../typings'
 import GraphQLError, { getErrorMessage } from '../../utils/GraphQLError'
 import checkConfig from '../config'
+import { organizationStatus } from '../fieldResolvers'
+
+export interface GetOrganizationsByEmailWithStatus {
```
enzomerca, line 15: sorry, I'm a bit confused. If this an interface, I think it can live in the `typings.ts` file (without the `export`) and include it here on the `import type { Organization } from '../../typings'` line (same as `Organization` that is also an interface)

enzomerca (dismissed):

PR #186, opened 2024-10-25
fix: check user permission error

Matheus-Aguilar:
This fix adds the behavior of avoiding the call of the `checkUserPermission` API of `storefront-permissions` when session data is not available. 

This API uses session data to get the user permissions, so any call without such data would result in an error in `storefront-permissions` (and consequently another one in `b2b-organizations-graphql`) and a `null` result. For this reason, it is not an issue to assume beforehand that the API call will return `null` when the session data is already empty.

Therefore, by avoiding this call when the session data is already known to be empty, the errors won't happen and an unnecessary call between apps will be prevented. The goal is to reduce the amount of error logs in both apps. More details on [https://vtex-dev.atlassian.net/browse/B2BTEAM-1689](https://vtex-dev.atlassian.net/browse/B2BTEAM-1689).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Fixed
 - Avoid calls to checkUserPermissions when session data is not available
 
 ## [0.61.0] - 2024-10-16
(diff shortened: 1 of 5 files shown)


enzomerca (dismissed):
Nice! Can you just update the changelog and also explain a bit more in the PR description why it is not an issue to not call the checkUserPermissions in case the session data is null?

enzomerca (approved):
It looks like we are repeating the same code in multiple places (get/check session and then get user permissions). I wonder if we could refactor this to reuse code.
I will let you decide if it is worth to try to refactor it or not. I'm ok to leave as it is if it refactoring is too much trouble.

PR #187, opened 2024-11-20
Mail notification to createOrganizationAndCostCenterWithAdminUser Mutation

arturmagalhaesjr:

There is no mail notification to `createOrganizationAndCostCenterWithAdminUser` mutation

It was added to that to keep the same behavior as others.

#### How to test it?

https://mail1--b2bstoreqa.myvtex.com/

### DIFF
--- a/node/resolvers/Mutations/Organizations.ts
+++ b/node/resolvers/Mutations/Organizations.ts
@@ -202,7 +202,7 @@ const createOrganizationAndCostCenterWithAdminUser = async (
   ctx: Context
 ) => {
   const {
-    clients: { storefrontPermissions },
+    clients: { storefrontPermissions, mail },
     vtex: { logger },
   } = ctx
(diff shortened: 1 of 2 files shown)


node/resolvers/Mutations/Organizations.ts
```diff
@@ -307,6 +307,12 @@ const createOrganizationAndCostCenterWithAdminUser = async (
       })
     }
 
+    message({
```
ataideverton, line 310: Do you think you should check the settings here and keep with up with the pattern for the other mutations?

Something like:


```suggestion
if(settings?.transactionEmailSettings?.organizationCreated) {
    message({
          logger,
          mail,
      storefrontPermissions,
    }).organizationCreated(organizationInput.name)
}
```

ataideverton (approved):

PR #203, opened 2025-07-23
fix: allow GetOrganizationRequestById to receive cookies

enzomerca:

`getOrganizationRequestById` requests were failing with `No auth cookie provided`.
This is because the cache control for this API was set to `PUBLIC`.
This PR changes the cache control for this API to `PRIVATE`. It also fixes the same thing for the `getCostCenters` APIs (that are less used, but would suffer from the same issue).


nicholas-maestrello (approved):
👍

PR #204, opened 2025-09-05
Feature/license manager checks

jorgeAcostaVTEX:

Added License Manager Permission check using Buyer Organizations for all queries and mutations (view/edit access). Admin and API tokens must have buyer_organization_view and/or buyer_organization_edit to use this app.

#### How to test it?

[jorgegraphql--vpcldev.myvtex.com](Link goes here!)

### DIFF
--- a/graphql/directives.graphql
+++ b/graphql/directives.graphql
@@ -1,6 +1,7 @@
 directive @withSession on FIELD | FIELD_DEFINITION
 directive @withPermissions on FIELD | FIELD_DEFINITION
 directive @validateAdminUserAccess on FIELD | FIELD_DEFINITION
+directive @validateAdminUserEditAccess on FIELD | FIELD_DEFINITION
 directive @validateStoreUserAccess on FIELD | FIELD_DEFINITION
 directive @checkUserAccess on FIELD | FIELD_DEFINITION
 directive @checkAdminAccess on FIELD | FIELD_DEFINITION
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -155,17 +155,17 @@ type Mutation {
     status: String!
     notes: String
     notifyUsers: Boolean
-  ): MutationResponse @validateAdminUserAccess
-  deleteOrganizationRequest(id: ID!): MutationResponse @validateAdminUserAccess
+  ): MutationResponse @validateAdminUserEditAccess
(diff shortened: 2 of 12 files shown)


graphql/schema.graphql
```diff
@@ -155,17 +155,20 @@ type Mutation {
     status: String!
     notes: String
     notifyUsers: Boolean
-  ): MutationResponse @validateAdminUserAccess
-  deleteOrganizationRequest(id: ID!): MutationResponse @validateAdminUserAccess
+  ): MutationResponse @validateAdminUserAccess(role: "B2B_ORGANIZATIONS_EDIT")
```
enzomerca, line 158: I know the default is `B2B_ORGANIZATION_VIEW`, but I wonder if we should add it to all APIs so it is explicit the required role for each API... 

Also, I think we will need to add the same parameter to the `validateStoreUserAccess` because we do have APIs with this directive that updates org/cost center.

enzomerca (commented):
Can we fix the lint errors as well?

graphql/schema.graphql
jorgeAcostaVTEX: HI @enzomerca! I am not sure I follow the guidance to add it to all APIs. Currently these roles are enforced only for the admin and API users (with the latest change) right?

graphql/directives.graphql
```diff
@@ -1,7 +1,11 @@
 directive @withSession on FIELD | FIELD_DEFINITION
 directive @withPermissions on FIELD | FIELD_DEFINITION
-directive @validateAdminUserAccess on FIELD | FIELD_DEFINITION
-directive @validateStoreUserAccess on FIELD | FIELD_DEFINITION
+directive @validateAdminUserAccess(
+  role: String = "B2B_ORGANIZATIONS_VIEW"
+) on FIELD | FIELD_DEFINITION
+directive @validateStoreUserAccess(
+  role: String = "B2B_ORGANIZATIONS_VIEW"
```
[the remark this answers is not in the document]
jorgeAcostaVTEX: Makes sense to me!

graphql/schema.graphql
jorgeAcostaVTEX: One question that came up during my testing, should users with B2B_ORGANIZATIONS_EDIT but no B2B_ORGANIZATIONS_VIEW be able to view organizations? Or should both permissions be explicit? @enzomerca

enzomerca (approved):

PR #206, opened 2025-09-15
Feature/audit mutations queries

jadso3145:
Audit logs were added to queries and mutations in the B2B Organizations GraphQL repository in order to improve traceability and change control.
These audits leverage the existing permission validation directives, so the audit flow is aligned with the same authorization checks already in place.

The motivation for this change is to add auditing to queries and mutations in the B2B Organizations GraphQL repository to improve traceability and accountability.

#### How to test it?
1) Link this branch to a Workspace.

2) Run the audited queries and mutations using the GraphQL IDE or verify them through the Admin Audit section.

3) Confirm that audit logs are generated correctly.

[Workspace](https://jaraudit--vpcldev.myvtex.com/)

These examples were taken from the GraphQL IDE.

successful
image

unauthorized
image

forbidden
image]

### DIFF
--- a/node/clients/audit.ts
+++ b/node/clients/audit.ts
@@ -1,0 +1,55 @@
+import type { InstanceOptions, IOContext } from '@vtex/api'
+import { ExternalClient } from '@vtex/api'
+
+export class AuditClient extends ExternalClient {
+  constructor(ctx: IOContext, options?: InstanceOptions) {
+    super('http://analytics.vtex.com', ctx, {
+      ...options,
+      headers: {
+        'Proxy-Authorization': ctx.adminUserAuthToken || '',
+        VtexIdClientAutCookie: ctx.adminUserAuthToken || '',
+        Authorization: ctx.authToken,
+        'X-Vtex-Use-Https': 'true',
+      },
+    })
+  }
+
+  public async sendEvent(
+    auditEntry: AuditEntry,
+    sessionMeta: any
+  ): Promise<void> {
+    console.log('sendEvent', auditEntry)
+    const { meta, subjectId, operation, authorId } = auditEntry
+    const { account, operationId, requestId, userAgent, logger } = this.context
+
+    console.log('sendEvent account', account)
+
+    const auditEvent = {
+      mainAccountName: account,
+      accountName: account,
+      id: requestId,
+      subjectId,
+      authorId,
+      application: 'vtex.b2b-organizations-graphql',
+      operation,
+      operationId,
+      meta: {
+        fowardFromVtexUserAgent: userAgent,
+        ...meta,
+      },
+      date: new Date().toJSON(),
+    }
+
+    console.log('auditEvent', auditEvent)
+
+    try {
+      await this.http.post(
+        `http://analytics.vtex.com/api/audit/events?an=${account}`,
+        auditEvent
+      )
+    } catch (error) {
+      console.log(sessionMeta, logger)
+      console.log('error ------', error)
+    }
+  }
+}
--- a/node/resolvers/Mutations/CostCenters.ts
+++ b/node/resolvers/Mutations/CostCenters.ts
@@ -35,11 +35,16 @@ const CostCenters = {
   ) => {
     const {
       vtex,
-      vtex: { logger },
+      vtex: { logger, adminUserAuthToken },
+      clients: { audit, licenseManager },
+      ip
     } = ctx
 
     // create schema if it doesn't exist
     await checkConfig(ctx)
+
+
+    const { profile } = await licenseManager.getTopbarData(adminUserAuthToken ?? '')
 
     if (!organizationId) {
       // get user's organization from session
@@ -74,12 +79,30 @@ const CostCenters = {
         paymentTerms,
       }
 
-      return await CostCenterRepository.createCostCenter(
+      const result = await CostCenterRepository.createCostCenter(
         _,
         organizationId,
         costCenter,
         ctx
       )
+
+      // Auditoría del acceso a createCostCenter
+      await audit.sendEvent({
+        subjectId: 'create-cost-center-event',
+        operation: 'CREATE_COST_CENTER',
+        authorId: profile?.id || 'unknown',
+        meta: {
+          entityName: 'CreateCostCenter',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({
+            organizationId,
+            costCenter
+          }),
+          entityAfterAction: JSON.stringify(result),
+        },
+      }, {})
+
+      return result
     } catch (error) {
       logger.error({
         error,
@@ -108,10 +131,20 @@ const CostCenters = {
     }: { organizationId: string; input: CostCenterInputWithId },
     ctx: Context
   ) => {
+    const {
+      vtex: { logger, adminUserAuthToken },
+      clients: { audit, licenseManager },
+      ip
+    } = ctx
+
     // create schema if it doesn't exist
     await checkConfig(ctx)
 
-    // check if organization exists
+
+    const { profile } = await licenseManager.getTopbarData(adminUserAuthToken ?? '')
+
+    try {
+      // check if organization exists
     const organization = (await Organizations.getOrganizationById(
       _,
       { id: organizationId },
@@ -158,7 +191,42 @@ const CostCenters = {
       ctx
     )
 
-    return { id: costCenterId }
+      await audit.sendEvent({
+        subjectId: 'create-cost-center-with-id-event',
+        operation: 'CREATE_COST_CENTER_WITH_ID',
+        authorId: profile?.id || 'unknown',
+        meta: {
+          entityName: 'CreateCostCenterWithId',
+          remoteIpAddress: ip,
+        entityBeforeAction: JSON.stringify({
+          organizationId,
+          input: {
+            id,
+            name,
+            addresses,
+            phoneNumber,
+            businessDocument,
+            stateRegistration,
+            customFields,
+            marketingTags,
+            sellers,
+            paymentTerms,
+          }
+        }),
+          entityAfterAction: JSON.stringify({
+            id: costCenterId
+          }),
+        },
+      }, {})
+
+      return { id: costCenterId }
+    } catch (error) {
+      logger.error({
+        error,
+        message: 'createCostCenterWithId-error',
+      })
+      throw new GraphQLError(getErrorMessage(error))
+    }
   },
 
   createCostCenterAddress: async (
@@ -167,12 +235,16 @@ const CostCenters = {
     ctx: Context
   ) => {
     const {
-      clients: { masterdata },
-      vtex: { logger },
+      clients: { masterdata, audit, licenseManager },
+      vtex: { logger, adminUserAuthToken },
+      ip
     } = ctx
 
     // create schema if it doesn't exist
     await checkConfig(ctx)
+
+
+    const { profile } = await licenseManager.getTopbarData(adminUserAuthToken ?? '')
 
     try {
       const costCenter: CostCenterInput = await masterdata.getDocument({
@@ -193,6 +265,24 @@ const CostCenters = {
         id: costCenterId,
       })
 
+      await audit.sendEvent({
+        subjectId: 'create-cost-center-address-event',
+        operation: 'CREATE_COST_CENTER_ADDRESS',
+        authorId: profile?.id || 'unknown',
+        meta: {
+          entityName: 'CreateCostCenterAddress',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({
+            costCenterId,
+            address
+          }),
+          entityAfterAction: JSON.stringify({
+            costCenterId,
+            status: 'success'
+          }),
+        },
+      }, {})
+
       return { status: 'success', message: '' }
     } catch (error) {
       logger.error({
@@ -205,14 +295,31 @@ const CostCenters = {
 
   deleteCostCenter: async (_: void, { id }: { id: string }, ctx: Context) => {
     const {
-      clients: { masterdata },
+      clients: { masterdata, audit, licenseManager },
+      vtex: { adminUserAuthToken },
+      ip
     } = ctx
+
+
+    const { profile } = await licenseManager.getTopbarData(adminUserAuthToken ?? '')
 
     try {
       await masterdata.deleteDocument({
         dataEntity: COST_CENTER_DATA_ENTITY,
         id,
       })
+
+      await audit.sendEvent({
+        subjectId: 'delete-cost-center-event',
+        operation: 'DELETE_COST_CENTER',
+        authorId: profile.id || '',
+        meta: {
+          entityName: 'DeleteCostCenter',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({ id }),
+          entityAfterAction: JSON.stringify({ deleted: true, id }),
+        },
+      }, {})
 
       return { status: 'success', message: '' }
     } catch (e) {
@@ -222,14 +329,31 @@ const CostCenters = {
 
   deleteOrganization: async (_: void, { id }: { id: string }, ctx: Context) => {
     const {
-      clients: { masterdata },
+      clients: { masterdata, audit, licenseManager },
+      vtex: { adminUserAuthToken },
+      ip
     } = ctx
+
+  
+    const { profile } = await licenseManager.getTopbarData(adminUserAuthToken ?? '')
 
     try {
       await masterdata.deleteDocument({
         dataEntity: ORGANIZATION_DATA_ENTITY,
         id,
       })
+
+      await audit.sendEvent({
+        subjectId: 'delete-organization-event',
+        operation: 'DELETE_ORGANIZATION',
+        authorId: profile.id || '',
+        meta: {
+          entityName: 'DeleteOrganization',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({ id }),
+          entityAfterAction: JSON.stringify({ deleted: true, id }),
+        },
+      }, {})
 
       return { status: 'success', message: '' }
     } catch (e) {
@@ -254,12 +378,16 @@ const CostCenters = {
     ctx: Context
   ) => {
     const {
-      clients: { masterdata },
-      vtex: { logger },
+      clients: { masterdata, audit, licenseManager },
+      vtex: { logger, adminUserAuthToken },
+      ip
     } = ctx
 
     // create schema if it doesn't exist
     await checkConfig(ctx)
+
+
+    const { profile } = await licenseManager.getTopbarData(adminUserAuthToken ?? '')
 
     try {
       await masterdata.updatePartialDocument({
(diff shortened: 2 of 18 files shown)


node/resolvers/Mutations/CostCenters.ts
```diff
@@ -35,12 +35,18 @@ const CostCenters = {
...
-      vtex: { logger },
+      vtex: { logger, adminUserAuthToken },
+      clients: { audit, licenseManager },
+      ip,
     } = ctx
 
     // create schema if it doesn't exist
     await checkConfig(ctx)
 
+    const { profile } = await licenseManager.getTopbarData(
```
enzomerca, line 46: it looks like we always do this call to get the user id, and we always provide the `adminUserAuthToken` (that goes on the lm request header). What if the API was called by a storefront user (where we don't have the `adminUserAuthToken`)? And what this `getTopbarData` returns exactly?

node/resolvers/Queries/Users.ts
```diff
@@ -229,7 +232,27 @@ const Users = {
       await vbase.saveJSON('b2borg', app, settings)
     }
 
-    return settings
+      // Auditoría del acceso a getAppSettings
```
enzomerca, line 235:
```suggestion
```

node/clients/audit.ts
```diff
@@ -0,0 +1,55 @@
...
+    const { account, operationId, requestId, userAgent, logger } = this.context
+
+    console.log('sendEvent account', account)
+
+    const auditEvent = {
+      mainAccountName: account,
+      accountName: account,
+      id: requestId,
+      subjectId,
+      authorId,
```
enzomerca, line 32: I wonder if the authorId and the ip address could be obtained inside this function. Mainly the authorId, so we don't need to add the `licenseManager.getTopbarData()` to a bunch of functions.

Or maybe we should clean this function and just keep the actual `post` call here, and move everything else (and the `getTopbarData` call) to a helper function somewhere else.

enzomerca (changes requested):
Could you clarify:
1/ Who is going to look at these audit logs? The merchant only?
2/ Do we want audit logs for both admin and storefront user operations? If both, do we want to identify the admin user and the storefront user?
3/ I looks like there is an audit app that already exists being used here. Which app exactly is that? (so I can install it in a test account)
4/ We want audit logs only for operations that succeeded, is that right? Do we need to audit `gets` as well? If so, why?

node/resolvers/Mutations/CostCenters.ts
```diff
@@ -74,12 +80,32 @@ const CostCenters = {
...
 
-      return await CostCenterRepository.createCostCenter(
+      const result = await CostCenterRepository.createCostCenter(
         _,
         organizationId,
         costCenter,
         ctx
       )
+
+      await audit.sendEvent(
```
enzomerca, line 90: what happens if this throws an exception? Are we going to fail the API call even though the operation succeed?

node/clients/audit.ts
```diff
@@ -0,0 +1,55 @@
...
+        'X-Vtex-Use-Https': 'true',
+      },
+    })
+  }
+
+  public async sendEvent(
+    auditEntry: AuditEntry,
+    sessionMeta: any
+  ): Promise<void> {
+    console.log('sendEvent', auditEntry)
```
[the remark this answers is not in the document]
jorgeAcostaVTEX: Hi @enzomerca!
Regarding your questions:
1. Yes, the merchant is the only one who would be auditing these logs via the admin, not the end user.
2. Yes this is a great poing! Both actions need to be logged, ideally with a distinction between storefront user and admin user. 
3. We are sending the events to the [vtex.admin-audit app ](https://github.com/vtex/admin-audit/tree/v2.23.1) whoich has been updated to receive these events from the b2b-organizations-graphql.
4. The client requires logs for attempts to alter or even view operations (so 401 or 403). They also require logs of queries to audit access to clinic information (as they handle controlled substances).

node/clients/audit.ts
jadso3145: Yes, that’s a good point. I’ll check it, I think I can get the authorId directly and remove the need to call licenseManager.getTopbarData() in each resolver.

node/resolvers/Mutations/CostCenters.ts
jadso3145: Would the right identifier for a storefront user be the userId from the sessionToken?.

From what I’ve seen, getTopbarData just gives us the profile (where we take the authorId), but that depends on having the adminUserAuthToken. What I could do is check if adminUserAuthToken is present, and if not, send the userId instead. assuming that’s what you mean by a storefront user.

node/resolvers/Queries/Users.ts
jadso3145: on it

node/resolvers/Queries/CostCenters.ts
```diff
@@ -101,6 +102,17 @@ const costCenters = {
...
       }
 
+      await audit.sendEvent({
+        subjectId: 'get-cost-center-by-id-event',
+        operation: 'GET_COST_CENTER_BY_ID',
+        meta: {
+          entityName: 'GetCostCenterById',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({ id }),
+          entityAfterAction: JSON.stringify({}),
```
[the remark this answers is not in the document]
jadso3145: We could remove what’s currently in entityBeforeAction without any problem, but I think it still provides a bit more context even if it’s just a query to know what’s being sent and to have at least some information visible in the audit-admin. But yes, it could be removed.

node/utils/audit.ts
```diff
@@ -0,0 +1,42 @@
...
+      operation: operation,
+      authorId: "unknown",
+      meta: {
+        entityName: entityNameFirstLetter,
+        remoteIpAddress: ip,
+        entityBeforeAction: JSON.stringify({}),
+        entityAfterAction: JSON.stringify({}),
+      },
+    })
+  }
```
enzomerca: Why do we need the `if/else if` here? In both cases we are doing the exact same call to `audit.sendEvent`

node/clients/audit.ts
```diff
@@ -0,0 +1,82 @@
...
+      authorId = await this.getUserIdFromSession(sessionToken)
+    }
+
+    const auditEvent = {
+      mainAccountName: account,
+      accountName: account,
+      id: requestId,
+      subjectId,
+      authorId,
+      application: 'vtex.b2b-organizations-graphql',
```
enzomerca: question: are we going to add audit logs in other apps as well? I suppose it might make sense to add something to the storefront-permissions and b2b-quotes-graphql apps maybe?

node/resolvers/directives/validateAdminUserAccess.ts
```diff
@@ -123,6 +124,16 @@ export class ValidateAdminUserAccess extends SchemaDirectiveVisitor {
...
       if (!hasAdminToken && !hasAdminTokenOnHeader && !hasApiToken) {
+        sendAuthMetric(
+          context,
+          logger,
+          new AuthMetric(
+            context?.vtex?.account,
+            metricFields,
+            'ValidateAdminUserAccessAudit'
+          ),
+        )
```
enzomerca: Can we remove the call to `sendAuthMetric` from here? We shouldn't add metrics in this PR, we should keep just the `audit` call.

```suggestion
```

node/resolvers/directives/validateAdminUserAccess.ts
```diff
@@ -131,6 +142,17 @@ export class ValidateAdminUserAccess extends SchemaDirectiveVisitor {
...
+
+      sendAuthMetric(
+        context,
+        logger,
+        new AuthMetric(
+          context?.vtex?.account,
+          metricFields,
+          'ValidateAdminUserAccessAudit'
+        ),
+      )
```
enzomerca: ditto: remove call to `sendAuthMetric`
```suggestion
```

node/resolvers/directives/validateStoreUserAccess.ts
```diff
@@ -155,6 +155,16 @@ export class ValidateStoreUserAccess extends SchemaDirectiveVisitor {
...
       if (!hasAdminToken && !hasApiToken && !hasStoreToken) {
+        sendAuthMetric(
+          context,
+          logger,
+          new AuthMetric(
+            context?.vtex?.account,
+            metricFields,
+            'ValidateStoreUserAccessAudit'
+          ),
+        )
```
enzomerca: ditto: remove `sendAuthMetric`

```suggestion
```

enzomerca (changes requested):

node/utils/audit.ts
jadso3145: > If we want to send an audit event only for status codes 401 and 403, we can put both checks inside the same if: `if (statusCode === 401 || statusCode === 403)`

Good call, I’ll update it

node/resolvers/Mutations/CostCenters.ts
```diff
@@ -329,6 +455,25 @@ const CostCenters = {
...
+            entityName: 'CostCenterAddress',
+            remoteIpAddress: ip,
+            entityBeforeAction: JSON.stringify({
+              costCenterId,
+              address,
+            }),
+            entityAfterAction: JSON.stringify({
+              costCenterId,
+              addresses,
+              status: 'success',
```
enzomerca: ditto. entityBeforeAction should have the original "addresses" (before editting it).

Also, I don't think it makes sense to add a "status: success" here. If we generated this event, it does mean that the update was successful.

node/resolvers/Mutations/CostCenters.ts
```diff
@@ -214,6 +286,18 @@ const CostCenters = {
...
 
+      await audit.sendEvent(
+        {
+          subjectId: 'delete-cost-center-event',
+          operation: 'DELETE_COST_CENTER',
+          meta: {
+            entityName: 'CostCenter',
+            remoteIpAddress: ip,
+            entityBeforeAction: JSON.stringify({ id }),
+            entityAfterAction: JSON.stringify({ deleted: true, id }),
```
enzomerca: ditto.
entityBeforeAction should contain the whole costcenter data (we must query MD first).
and the entityAfterAction should be null/empty.

Also, I don't think it makes sense to add a "deleted: true" here. If we generated this event, it does mean that the delete was successful.

node/resolvers/Mutations/MarketingTags.ts
```diff
@@ -20,6 +21,23 @@ const MarketingTags = {
...
+          entityName: 'MarketingTags',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({
+            costId,
+            tags
+          }),
+          entityAfterAction: JSON.stringify({
+            costId,
+            tags
+          }),
```
enzomerca: ditto.
It doesn't make sense to have the entityBeforeAction the same as the entityAfterAction. Lets query vbase to get the entityBeforeAction data.

node/resolvers/Mutations/Organizations.ts
```diff
@@ -382,6 +383,27 @@ const Organizations = {
...
+            costCenters,
+            customFields,
+            paymentTerms,
+            priceTables,
+            salesChannel,
+            sellers,
+          }),
+          entityAfterAction: JSON.stringify(createOrganizationResult),
+        },
+      })
```
enzomerca: ditto.
entityBeforeAction should be null/empty.
entityAfterAction should have the new organization data.

node/resolvers/Mutations/Organizations.ts
```diff
@@ -610,6 +672,22 @@ const Organizations = {
...
+        meta: {
+          entityName: 'OrganizationRequest',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({
+            id
+          }),
+          entityAfterAction: JSON.stringify({
+            id,
+            deleted: true
+          }),
```
enzomerca: ditto.
entityBeforeAction should have the organization request data. entityAfterAction should be null/empty.

Also there is no need to add the "deleted: true". If this event was generated, it means the delete succeeded.

node/resolvers/Mutations/Organizations.ts
```diff
@@ -844,6 +963,26 @@ const Organizations = {
...
+            id,
+            status: organizationRequest.status,
+            notes: '',
+            organizationRequest: organizationRequest
+          }),
+          entityAfterAction: JSON.stringify({
+            id,
+            status,
+            notes
+          }),
```
enzomerca: ditto. same as the organization request approved case.

node/resolvers/Mutations/Users.ts
```diff
@@ -358,14 +373,27 @@ const Users = {
...
     )
+
+    await audit.sendEvent({
+      subjectId: 'impersonate-user-event',
+      operation: 'IMPERSONATE_USER',
+      meta: {
+        entityName: 'User',
+        remoteIpAddress: ip,
+        entityBeforeAction: JSON.stringify({ clId, userId }),
+        entityAfterAction: JSON.stringify(result),
```
enzomerca: ditto. Maybe the entityBeforeAction should be null/empty and the entityAfterAction should have the impersonated user data (the `userData` object maybe?)

node/resolvers/Mutations/Users.ts
```diff
@@ -249,11 +253,20 @@ const Users = {
...
-        status: 'error',
-      }
-    )
+    const result = impersonation?.data?.impersonateUser ?? { status: 'error' }
+
+    await audit.sendEvent({
+      subjectId: 'impersonate-b2b-user-event',
+      operation: 'IMPERSONATE_B2B_USER',
+      meta: {
+        entityName: 'B2BUser',
```
enzomerca: I think we are always talking about an "User" from an organization here, so maybe lets keep the entity just as "User" across the app in cases like this.

```suggestion
        entityName: 'User',
```

node/resolvers/Mutations/Users.ts
```diff
@@ -401,6 +434,22 @@ const Users = {
...
+              meta: {
+                entityName: 'RemoveUserWithEmail',
+                remoteIpAddress: ip,
+                entityBeforeAction: JSON.stringify({
+                  orgId,
+                  costId,
+                  email,
+                  user
+                }),
+                entityAfterAction: JSON.stringify(response.data.deleteUser),
```
enzomerca: entityAfterAction should be null/empty

node/resolvers/Mutations/Users.ts
```diff
@@ -542,11 +617,32 @@ const Users = {
...
+              roleId,
+              userId,
+              orgId,
+              costId,
+              clId,
+              name,
+              email
+            }),
+            entityAfterAction: JSON.stringify(result.data.addUser),
+          },
```
enzomerca: ditto.
entityBeforeAction should be null/empty
entityAfterAction should contain only the added user data

node/resolvers/Mutations/Users.ts
```diff
@@ -589,6 +691,24 @@ const Users = {
         canImpersonate,
       })
 
+      await audit.sendEvent({
+        subjectId: 'create-user-with-email-event',
+        operation: 'CREATE_USER_WITH_EMAIL',
+        meta: {
+          entityName: 'Email',
```
enzomerca:
```suggestion
          entityName: 'User',
```

node/resolvers/Mutations/Users.ts
```diff
@@ -667,12 +791,32 @@ const Users = {
...
+              id,
+              roleId,
+              userId,
+              orgId,
+              costId,
+              clId,
+              name,
+              email
+            }),
+            entityAfterAction: JSON.stringify(result.data.updateUser),
```
enzomerca: ditto.
we must query the user current user data to add to the entityBeforeAction, and add the new user data to the entityAfterAction.

node/resolvers/Queries/Organizations.ts
```diff
@@ -396,6 +460,18 @@ const Organizations = {
     }
 
     try {
+
+      await audit.sendEvent({
+        subjectId: 'get-organization-by-id-storefront-event',
+        operation: 'GET_ORGANIZATION_BY_ID_STOREFRONT',
+        meta: {
+          entityName: 'OrganizationByIdStorefront',
```
enzomerca:
```suggestion
          entityName: 'Organization',
```

node/resolvers/Queries/Settings.ts
```diff
@@ -78,18 +116,33 @@ const B2BSettings = {
...
+        operation: 'GET_ACCOUNT',
+        meta: {
+          entityName: 'Account',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify({}),
+          entityAfterAction: JSON.stringify({}),
+        },
+      })
+
+      return await lm.getAccount()
```
enzomerca: this way we won't emit the event if the get failed.
```suggestion
      const result = await lm.getAccount()
      await audit.sendEvent({
        subjectId: 'get-account-event',
        operation: 'GET_ACCOUNT',
        meta: {
          entityName: 'Account',
          remoteIpAddress: ip,
          entityBeforeAction: JSON.stringify({}),
          entityAfterAction: JSON.stringify({}),
        },
      })

      return result
```

node/resolvers/Queries/Users.ts
```diff
@@ -434,9 +472,21 @@ const Users = {
     return storefrontPermissions
       .listUsersPaginated(variables)
       .then((result: any) => {
-        return result.data.listUsersPaginated
+
+        audit.sendEvent({
+          subjectId: 'get-users-paginated-event',
+          operation: 'GET_USERS_PAGINATED',
+          meta: {
+            entityName: 'GetUsersPaginated',
```
enzomerca:
```suggestion
            entityName: 'Users',
```

node/utils/metrics/auth.ts
```diff
@@ -42,7 +42,7 @@ export class AuthMetric implements Metric {
 const sendAuthMetric = async (
   ctx: Context,
   logger: Logger,
-  authMetric: AuthMetric
+  authMetric: AuthMetric,
```
enzomerca:
```suggestion
  authMetric: AuthMetric
```

enzomerca (changes requested):
Please, after fixing the requested changes, mark the comments as resolved only after double checking in the diff if the comment was actually resolved.

node/resolvers/Mutations/Organizations.ts
```diff
@@ -582,6 +629,20 @@ const Organizations = {
...
+        meta: {
+          entityName: 'OrganizationAndCostCentersWithId',
+          remoteIpAddress: ip,
+          entityBeforeAction: JSON.stringify(input),
+          entityAfterAction: JSON.stringify({
+            href,
+            id
+          }),
+        },
+      })
```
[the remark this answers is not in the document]
jadso3145: I’ve updated the implementation accordingly.
For entityBeforeAction, I used JSON.stringify(null) instead of plain null to match the expected string type from the audit client.

node/resolvers/Mutations/Users.ts
jadso3145: @enzomerca, I suggest keeping `result` in `entityAfterAction`. The reason is that `result` represents the actual outcome of the impersonation operation, including success or failure. Using `userData` would only capture the intended user object, but it wouldn’t reflect if the impersonation actually succeeded or if there were any errors. From an auditing perspective, logging `result` provides a more accurate and complete record of what happened.

node/resolvers/Queries/CostCenters.ts
```diff
@@ -256,14 +308,27 @@ const costCenters = {
...
     // create schema if it doesn't exist
     await checkConfig(ctx)
 
     try {
+
+      await audit.sendEvent({
+        subjectId: 'get-cost-centers-by-organization-id-event',
+        operation: 'GET_COST_CENTERS_BY_ORGANIZATION_ID',
+        meta: {
+          entityName: 'CostCentersByOrganizationId',
```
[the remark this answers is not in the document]
enzomerca: This still incorrect and need to be changed to 'CostCenters'

nicholas-maestrello (approved):

PR #209, opened 2025-10-02
New pr with changelog and policies reconciled

jorgeAcostaVTEX:

Added buyer organization view and edit policies and changelog data to this new versino

#### How to test it?

[Workspace](Link goes here!)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "b2b-organizations-graphql",
   "vendor": "vtex",
-  "version": "2.0.0",
+  "version": "1.1.0",
   "title": "B2B Organizations",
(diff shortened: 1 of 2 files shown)


manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "b2b-organizations-graphql",
   "vendor": "vtex",
-  "version": "2.0.0",
+  "version": "1.1.0",
```
enzomerca, line 4: lets not touch this version. this should be done only during the release process, which is made automatically when merging the PR.
```suggestion
  "version": "2.0.0",
```

enzomerca (changes requested):

PR #223, opened 2026-05-27
[B2BTEAM-3516] Fixed collections being set as null after update

wender:

Partners integrating with B2B Organizations hit an inconsistency between **`createOrganizationAndCostCentersWithId`** (which accepts collection **names** via `NormalizedOrganizationInput` and resolves Catalog `{ id, name }`) and **`updateOrganization`** (which persisted `CollectionInput` as-is). Payloads with only `{ name }` stored collections **without a valid `id`**, breaking storefront catalog behavior after an update.

This change **resolves name-only `CollectionInput` rows** on update using the same Catalog lookup as create, **keeps full backward compatibility** when `id` is provided, and **only writes `collections` to Master Data when the argument is explicitly sent** so omitted `collections` no longer risk clearing existing data. **`findCollections`** also filters out unresolved entries after lookup instead of leaving nulls in the array.

GraphQL schema docs were updated for `updateOrganization` and `CollectionInput`; **CHANGELOG** updated under Unreleased.

---

#### How to test it?

- [ ] Link a **development workspace** where this app version is linked (e.g. `vtex link` on this branch).
- [ ] **Create** an org with `createOrganizationAndCostCentersWithId` using `collections: ["<existing collection name>"]`.
- [ ] **Query** `getOrganizationById` and confirm `collections` have both `id` and `name`.
- [ ] **Update** the same org with `updateOrganization`, passing `collections: [{ name: "<same name>" }]` only (no `id`); query again and confirm **`id` is still present** and matches Catalog.
- [ ] **Update** with `collections: [{ id: "...", name: "..." }]` and confirm behavior unchanged for existing integrations.
- [ ] **Update** without the `collections` argument and confirm **existing collections are unchanged** in Master Data.
- [ ] (Optional) Run unit tests: `cd node && yarn test Organizations.test.ts`.

**Workspace link:** _&lt;add your workspace URL here, e.g. `https://workspace--account.myvtex.com/admin/graphql-ide` 

Result before the fix:
image

Result after the fix
image

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "b2b-organizations-graphql",
   "vendor": "vtex",
-  "version": "2.4.4",
+  "version": "2.4.5-beta.0",
   "title": "B2B Organizations",
   "description": "App to create and manage B2B Organizations and Cost Centers",
   "mustUpdateAt": "2022-08-28",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`updateOrganization` collections**: resolve `CollectionInput` entries by Catalog name when no `id` is sent, matching `createOrganizationAndCostCentersWithId`/`NormalizedOrganizationInput` behavior so storefront catalog rules stay valid after updates.
+- **Partial updates**: only send `collections` to Master Data when the argument is explicitly provided (`undefined`/omitted leaves existing collections unchanged).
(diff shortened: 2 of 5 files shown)


manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "b2b-organizations-graphql",
   "vendor": "vtex",
-  "version": "2.4.4",
+  "version": "2.4.5-beta.0",
```
enzomerca, line 4: só lembrar de não commitar esse arquivo junto!

enzomerca (approved):

renan-vtex (approved):

manifest.json
wender: Ele ajusta no versionamento

PR #229, opened 2026-07-21
Fix/b2 bteam 3729 async audit event

wender:

Two related problems, both traced back to Master Data schema versioning.

**1. Intermittent login timeouts** (Zendesk #1433601, B2BTEAM-3729, go-live blocker)

The `session transform` (`setProfile`, in `vtex.storefront-permissions`) calls `getCostCenterById`, `getMarketingTags` and `getB2BSettings` from this app in parallel, under a 5s client timeout. Users occasionally got `ECONNABORTED`, failing the whole session transform and blocking login.

`AuditClient.sendEvent` (called from all three queries) synchronously awaited a full session fetch (`session.getSession(sessionToken, ['*'])`, added in 53a8045) plus an outbound HTTP POST to `analytics.vtex.com`. In `getMarketingTags` this even ran *before* the actual data fetch. So every login paid for 3x redundant full-session reads plus 3x calls to the audit service, inside the caller's 5s budget. Any latency on the session service or analytics.vtex.com — likelier while the session itself is mid-transform — pushed all three past the timeout at once, matching the intermittent pattern reported.

> Note: this addresses one of three contributors identified. Still pending (not in this PR): `checkConfig` running a vbase read on every request, and the write-on-read in `addMissingAddressIds`. The beta was still able to reproduce the timeout, which is expected for that reason.

**2. `403 Unauthorized Access` for some users on storefront B2B flows**

Reported after installing `2.6.1-beta.0` on `kohlerqa`/`kohlerdev`: `getOrganizationByIdStorefront`, `getCostCenterByIdStorefront` and `getActiveOrganizationsByEmail` returned `ForbiddenError` — but only for a handful of users.

Root cause: `USER_DATA_ENTITY_SCHEMA` was bumped from `v0.1.2` to `v3.1.2` in 8013365 ("Updating schema version"), which shipped in `2.6.1-beta.0` and is not in `2.5.1-hkignore.0`. The `b2b_users` documents in the affected accounts are registered under `v0.1.2`, so the search in `isUserPartOfBuyerOrg` returned no rows for users whose documents are not associated with the newer schema version. That made `isUserPartOfBuyerOrg` report `false`, `validateStoreToken` left `hasValidStoreToken` false, and `@validateStoreUserAccess` answered 403.

This explains both puzzling symptoms: why it hit **only some users** (only those whose documents weren't associated with `v3.1.2`), and why **rolling back to `2.5.1-hkignore.0` made it disappear** (that version searches with `v0.1.2`). Confirmed against the OpenSearch log `"Invalid store token:"`, which is emitted precisely when the VTEX ID token is valid but the buyer-org check fails.

The `mdSchema.ts` schema names (`ORGANIZATION_*`, `COST_CENTER_SCHEMA_VERSION`) were held at `v0.x` for the same reason — that part was split out and merged via #230; this branch carries it.

#### How to test it?

**Timeout / audit**
1. Log in on a storefront where this app is installed and confirm `setProfile`/session transform responds well under 5s, even when `analytics.vtex.com` is slow or unreachable (login must still succeed — audit failures are logged, never surfaced to the caller).
2. Confirm audit events for `getCostCenterById`, `getMarketingTags`, `getB2BSettings` and the other call sites still reach the audit trail (`/api/audit/events`) under normal conditions — the event payload is unchanged, only its delivery is now detached.

**403 / store user authorization**
3. With a user that was previously getting the 403, confirm `getOrganizationByIdStorefront`, `getCostCenterByIdStorefront` and `getActiveOrganizationsForSelector` now resolve normally.
4. Confirm `USER_DATA_ENTITY_SCHEMA` matches the schema the account's `b2b_users` documents are actually registered under:
   `GET /api/dataentities/b2b_users/search?_fields=id,email,schemas&_schema=v0.1.2&_where=(email=<user>)`
5. In OpenSearch, confirm the new diagnostics appear when a user is *not* found: `isUserPartOfBuyerOrg.searchResult` (schema, filter, result count), `validateStoreToken.buyerOrgCheck` (buyer-org outcome), and `isUserPartOfBuyerOrg.searchError` for the Master Data error that was previously swallowed by an empty `catch`.

N/A — backend fix, no UI change. See the HAR files, cURL reproduction and OpenSearch export shared in the incident thread.

For the timeout: considered raising the client timeout or adding retry/backoff on the storefront-permissions side. Chose to make audit fire-and-forget because it removes genuinely unnecessary work from the login-critical path instead of giving it more time to fail slowly.

For the 403: the real fix is to make the schema version and the documents' registered version move together. Holding `v0.1.2` restores the working state; any future roll-forward to major-based schemas needs `mdSchema.ts` **and** `USER_DATA_ENTITY_SCHEMA` handled as one coordinated migration, since these two incidents came from splitting them.

#### Related to / Depends on

- B2BTEAM-3729 / Zendesk #1433601 (Kohler login timeouts + 403)
- Reverts the schema version bumps from 62388ba / B2BTEAM-3598 (major-based schemas) and 8013365
- #230 (mdSchema schema-name hold, already merged)


nicholas-maestrello (approved):
LGTM

pabloppupulin (approved):

